PR #2, opened 2019-11-11
Initial implementation of HearingResult model with basic specs

archferns:
Add Hearing Model to store the `body` of the `hearing/results?hearingId=XX` endpoint.
If the Hearing does not exist, creates a new record, or else updates it.

### DIFF
--- a/app/services/hearing_result_fetcher.rb
+++ b/app/services/hearing_result_fetcher.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class HearingResultFetcher < ApplicationService
+  def initialize(hearing_id)
+    @hearing_id = hearing_id
+  end
+
+  def call
+    connection.get("/hearing/results/#{hearing_id}")
+  end
+
+  private
+
+  def connection
+    @connection ||= Faraday.new ENV.fetch('COMMON_PLATFORM_URL') do |conn|
+      conn.response :json, content_type: 'application/json'
+      conn.adapter Faraday.default_adapter
+    end
+  end
+
+  attr_reader :hearing_id
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,0 +1,25 @@
+# This file is auto-generated from the current state of the database. Instead
+# of editing this file, please use the migrations feature of Active Record to
+# incrementally modify your database, and then regenerate this schema definition.
+#
+# This file is the source Rails uses to define your schema when running `rails
+# db:schema:load`. When creating a new database, `rails db:schema:load` tends to
+# be faster and is potentially less error prone than running all of your
+# migrations from scratch. Old migrations may fail to apply correctly if those
+# migrations use external dependencies or application code.
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 2019_11_11_162455) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "pgcrypto"
+  enable_extension "plpgsql"
+
+  create_table "hearings", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.jsonb "body", null: false
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+  end
+
+end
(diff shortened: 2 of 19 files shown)


app/models/hearing_result.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class HearingResult < ApplicationRecord
+  validates :hearing_id, presence: true
+  validates :response, presence: true
```
scruti: By calling this `response` aren't we binding it to only the Search/Response endpoint? This model is also used for the hearing result updates pushed from the Common Platform without a search right? 
What about just calling it `hearing`?

db/schema.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 2019_11_11_162455) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "pgcrypto"
+  enable_extension "plpgsql"
+
+  create_table "hearing_results", force: :cascade do |t|
```
scruti: Open question 💭 

As, if I remember right, we will be grouping the hearings by their Prosecution Case, would make sense to have a `ProsecutionCaseID` indexed column (populated from the JSON values when populating the model) so we can create rails associations between the `ProsecutionCases` and the `HearingResults`?  Or leave it like this, create a JsonB index on the `ProsecutionCaseID` attribute in the JSONB and directly create associationgs with a gem like this: https://github.com/y9v/activerecord-jsonb-associations ?
Or any other options?

app/models/hearing_result.rb
archferns: Both the [search](https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/master/lib/schemas/api/results.hearingResultedResponse.json) and [publish](https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/master/lib/schemas/api/hearing-resulted.json) responses have a `hearing` and `sharedTime`, hence I chose to call it `HearingResult` rather than just `Hearing`, but i'm happy to change it to `Hearing`
About `response`, in the context of a request/response rather than search/response. Would it make sense to have it called something like `body` (I've assumed we are storing the response body only for a successful response, should we be concerned with the response status, headers etc?)

db/schema.rb
archferns: I think this will be the other way around, because hearings have a one to many relationship with [prosecution cases](https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/master/lib/schemas/global/hearing.json#L21-L28)
We can add a hearing_id on the `prosecution_cases` table I suppose?

app/models/hearing_result.rb
scruti: No problem with the `HearingResult` (as can be read as some sort of output from a hearing). I would be happy either way `Hearing` or `HearingResult`.

On the `response`  I like your `body` suggestion :clap:. Good name.
Was concerned about the `HearingResulted` coming from HMCCTS being stored here too and `response` seems out of context in that case. But `body` seems perfect! :)

db/schema.rb
scruti: Been checking/refreshing on the endpoint data views and the json schemas.

From the initial search (that returns a Prosecution Case with its associated hearings) to the Record LAA Reference (over a Prosecution Case) seems the driver for collecting and searching the hearings will be the `ProsecutionCaseID`
Then, when we get the hearing, as you said, seems a hearing can be associated to multiple prosecution cases. 
We definitely need to be able to have an association that allows us for a quick "give me all the stored hearings for this ProsecutionCase".

Maybe a many to [many association](https://guides.rubyonrails.org/association_basics.html#the-has-and-belongs-to-many-association) is the way to go?

app/services/hearing_result_fetcher.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class HearingResultFetcher < ApplicationService
```
scruti: As we renamed the `HearingResult`model to `Hearing`, would make sense to normalize it into the services too so we have `HearingFetcher` and `HearingRecorder` ?

archferns: Oops, sorry, did that retrospectively https://github.com/ministryofjustice/laa-court-data-adaptor/pull/4

PR #5, opened 2019-11-14
Add POST endpoint to receive hearings

archferns:
Endpoint for Publish Hearing-Resulted API.

Added a service which can be triggered via a background job/rake task depending on the use case as below:
```ruby
Api::GetHearingResults.call(hearing_id)
```
This will call the mock endpoint and record the body of the response against the `Hearing`.

### DIFF
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -4,8 +4,8 @@ class HearingsController < ApplicationController
   def create
     @hearing = HearingRecorder.call(params[:hearing][:id], hearing_params)
 
-    if @hearing
-      render json: @hearing, status: :created
+    if @hearing.valid?
+      head :created
     else
       render json: @hearing.errors, status: :unprocessable_entity
     end
--- a/app/controllers/hearings_controller.rb
+++ b/app/controllers/hearings_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class HearingsController < ApplicationController
+  def create
+    @hearing = HearingRecorder.call(params[:hearing][:id], hearing_params)
+
+    if @hearing
+      render json: @hearing, status: :created
+    else
+      render json: @hearing.errors, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  # Only allow a trusted parameter "white list" through.
+  def hearing_params
+    params.require(%i[sharedTime hearing])
+    params.permit(:sharedTime, hearing: {})
+  end
+end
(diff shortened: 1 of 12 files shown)


app/controllers/hearings_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class HearingsController < ApplicationController
+  def create
+    @hearing = HearingRecorder.call(params[:hearing][:id], hearing_params)
+
+    if @hearing
+      render json: @hearing, status: :created
```
archferns: HMCTS expect a `location` header, however, this is omitted, since we do not have any provision for querying via the adaptor yet.
https://hmcts.github.io/restful-api-standards/#success-codes

scruti: Really good catch! Going to add this as a ticket in the backlog so we don't forget about this bit.

scruti: Also seems optional right? `You are free to return either an empty response or the created resource in conjunction with the Location header`

archferns: It seemed a bit ambiguous, i.e , does it mean they need (both response + location) or (no response but _always_ location)
However, probably it is best to return an empty response, since otherwise we have to consider further questions about serialisation.

app/services/hearing_recorder.rb
```diff
@@ -1,17 +1,17 @@
 # frozen_string_literal: true
 
 class HearingRecorder < ApplicationService
-  def initialize(hearing_id)
+  def initialize(hearing_id, body = nil)
     @hearing = Hearing.find_or_initialize_by(id: hearing_id)
+    @body = body.presence || HearingFetcher.call(hearing.id).body
```
scruti: Smart 👏

scruti: Seems this is a good point to extract `HearingFetcher` call outside this class and inject the body either coming from the `HearingUpdated` update from HMCTS or through `HearingFetcher` call.

Should we invert the dependency between these two services so we call `HearingRecorder` from within `HearingFetcher` call? 
In the near future probably we will have a controller action calling both `HearingFetcher` and `HearingRecorder` together so the services are not coupled between them. But as a middle step inverting the dependency so `HearingRecorder` codebase does not have conditional flows and gets called the same from both `HearingFetcher` and from the `HearingsController` seems a good compromise meanwhile.

Let me know your thoughts on this :)

app/controllers/hearings_controller.rb
scruti: Left the spike in the backlog anyways so we can be sure we will double check it with more time. (Even if it is discarded in 10 minutes :D )

app/services/hearing_recorder.rb
archferns: Definitely agree on that, it makes sense to invert them. This version doesn't make sense with the POST endpoint addition.

PR #8, opened 2019-11-19
Add RecordLaaReference functionality

archferns:
[Creates/Updates](https://hmcts.github.io/restful-api-standards/#put) an `LaaReference` against an `Offence`.

### DIFF
--- a/app/services/api/record_laa_reference.rb
+++ b/app/services/api/record_laa_reference.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Api
+  class RecordLaaReference < ApplicationService
+    include CommonPlatformConnection
+
+    # rubocop:disable Metrics/ParameterLists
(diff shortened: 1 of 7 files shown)


app/services/api/record_laa_reference.rb
```diff
@@ -0,0 +1,38 @@
...
+        defendantId: defendant_id,
+        offenceId: offence_id,
+        statusCode: status_code,
+        applicationReference: application_reference,
+        statusDate: status_date
+      }
+    end
+
+    attr_reader :url, :prosecution_case_id, :defendant_id, :offence_id, :status_code, :application_reference, :status_date, :response
+  end
```
willmcb: Where is this response variable being declared? I get that the `CommonPlatformConnection` has a response - but I can't see how this would be accessible to an `attr_reader` on `RecordLaaReference`.

archferns: Good catch, this isn't needed/used here at all. Removed it.

PR #9, opened 2019-11-19
Remove unused railties

archferns:
Remove unused railties: `ActiveStorage`, `ActionMailer`, `ActionMailbox`, `ActionText` and minor rails version bump


config/application.rb
```diff
@@ -7,11 +7,11 @@
 require 'active_model/railtie'
 require 'active_job/railtie'
 require 'active_record/railtie'
-require 'active_storage/engine'
+# require "active_storage/engine"
 require 'action_controller/railtie'
-require 'action_mailer/railtie'
-require 'action_mailbox/engine'
-require 'action_text/engine'
+# require "action_mailer/railtie"
```
willmcb: Why comment them out rather than remove them? Do you think we might need them in the future?

archferns: This was generated via the normal generator:
`rails new laa-court-data-adaptor --skip-action-mailbox -d postgresql --skip-test --skip-action-mailer --skip-active-storage --api`
I doubt we would need these, but it is hopefully useful for more context.

PR #11, opened 2019-11-20
LASB-235 - implement search case

scruti:
This is work in progress for consuming the `Search Prosecution Case` endpoint in HMCTS Cloud Platform API.

The current code covers the search by `prosecutionCaseReference`. Further work needs to be done to extend the codebase in order to allow to search the endpoint based in other parameters:
Screen Shot 2019-11-20 at 17 20 57


scruti:
Thanks @archferns and @willmcb for take over this :clap:

PR #18, opened 2019-12-04
Add ProsecutionCase#index endpoint

archferns:
Add endpoint for MLRA to communicate to the adaptor


config/routes.rb
```diff
@@ -4,4 +4,6 @@
   # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
   resources :status, only: [:index]
   resources :hearings, only: [:create]
+  resources :prosecution_cases, only: [:index]
+  post 'prosecution_cases' => 'prosecution_cases#index'
```
archferns: Do we want this to be a post?

PR #22, opened 2019-12-16
Add user authentication

archferns:
Adds a User model 
Add Authorisation via HTTP Token auth to the controller.

The expected format for this is as follows:
```
  Authorization: Bearer SECRETTOKEN, user_id=USER-UUID
```
This allows us to fetch the user by the id and then verify by the token
and avoids the need for storing the `SECRETTOKEN`
in plaintext or encrypted, since it is stored as a BCrypt digest
using the standard rails `has_secure_password` method.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::API
+  include ActionController::HttpAuthentication::Token::ControllerMethods
+  before_action :authenticate
+
   ERROR_MAPPINGS = {
     ActionController::ParameterMissing => :bad_request
   }.freeze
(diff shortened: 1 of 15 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ApplicationController, type: :controller do
```
scruti: Very nicely tested 👏

app/controllers/application_controller.rb
```diff
@@ -10,4 +13,15 @@ class ApplicationController < ActionController::API
       render json: { error: error }, status: status
     end
   end
+
+  private
+
+  def authenticate
+    authenticate_with_http_token do |token, options|
+      user = User.find_by(id: options[:user_id])
+      @current_user = user&.authenticate_auth_token(token)
```
scruti: Small question about the variable scope:
does `@current_user` needs to be an instance var in our service? 
I know it tends to be the default but I suspect we only need it  in `authenticate` method scope and don't use it anywhere else...

archferns: Yep, that is very good point about it not being needed outside the scope of this check. Ill refactor it.

spec/controllers/application_controller_spec.rb
archferns: ty :bowing_man:

PR #54, opened 2020-01-30
Use jsonapi schema format.

archferns:
## What

Modify schema to use [jsonapi](https://jsonapi.org/) format.

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-79)

## Checklist


schema/schemata/prosecution_case.json
```diff
@@ -5,16 +5,57 @@
...
+        }
+      }
+    },
     "id": {
       "description": "Unique identifier of prosecution case provided by HMCTS (prosecutionCaseId)",
       "readOnly": true,
       "format": "uuid",
       "type": "string"
     },
+    "type": {
```
jsugarman: are there going to be any other types than `prosecution_cases`?

archferns: We would probably have `hearings`, `defendants`, and others defined at their other endpoints, but `type` is a reserved jsonapi keyword to identify the resource type: https://jsonapi.org/format/#document-resource-objects

Issue #55, opened 2020-01-31
Test hyper-schema


jsugarman:
The available schema testing gems include
[json-schema](https://github.com/ruby-json-schema/json-schema) - does not support `hyper-schema`
[json_schemer](https://github.com/davishmcclurg/json_schemer) - does not support `hyper-schema`

[json_schema](https://github.com/brandur/json_schema) - supports `v4 hyper-schema` but raises an error against the current `schema.json`

```
schema_data = JSON.parse(File.read("schema.json"))
schema = JsonSchema.parse!(schema_data)
=> JsonSchema::AggregateError
```
I am not sure if this is a problem or not but should probably be looked into if we are aiming for [jsonapi](https://jsonapi.org) compliant responses

archferns:
Fixed via #56

PR #56, opened 2020-01-31
Switch to json-schema instead of hyper-schema

archferns:
## What

[Link to story](https://github.com/ministryofjustice/laa-court-data-adaptor/issues/55)
Change to json-schema to allow simplify validation with https://github.com/ruby-json-schema/json-schema

## Checklist


archferns:
Yep, sounds fine to me, I'm not sure if we need links anyway at the moment

PR #61, opened 2020-02-05
Expose endpoint to search prosecution_cases

archferns:
## What
Build upon the agreed schema for the UI to expose a JSONAPI compliant endpoint to query prosecution cases.

## Checklist


spec/requests/api/prosecution_cases_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
jsugarman: if you want to remove the need to require this you can amend [.rspec](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/.rspec) with this
```
--require rails_helper
```

spec/controllers/api/prosecution_cases_controller_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::ProsecutionCasesController, type: :controller do
```
jsugarman: **not a blocker**. I avoid controller specs generally and i see you have request specs that [cover this here](https://github.com/ministryofjustice/laa-court-data-adaptor/pull/61/files#diff-e01c816e3c08acaf4883309f914d0904)

spec/services/api/search_prosecution_case_spec.rb
```diff
@@ -20,6 +20,10 @@
     subject
   end
 
+  it 'returns the recorded ProsecutionCases' do
+    expect(subject).to all(be_a(ProsecutionCase))
```
jsugarman: could use `is_expected`  - same as `expect(subject)`

spec/requests/api/prosecution_cases_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  describe 'GET /api/prosecution_cases' do
+    around do |example|
+      VCR.use_cassette('search_prosecution_case/by_prosecution_case_reference_success') do
+        example.run
+      end
+    end
+
+    let(:valid_query_params) { { filter: { prosecution_case_reference: 'TFL12345' } } }
+    let(:schema) { File.read('schema/schema.json') }
+
```
jsugarman: possible refactor
```
before do
  get api_prosecution_cases_path, params: valid_query_params
end

it 'successful' do
  expect(response).to have_http_status(200)
end

 it 'matches the given schema' do
   valid = JSON::Validator.validate!(schema, response.body, fragment: '#/definitions/prosecution_case/definitions/resource_collection', strict: true)
    expect(valid).to be_truthy
end
```

I have written a little [RSpec matcher to cover validating](https://github.com/ministryofjustice/laa-court-data-ui/compare/add-mock-adaptor-data#diff-b9bae26c3443ee588ca9448022dd04d9) if you like

archferns: Thanks, I like this matcher! will pull it into this codebase.

spec/controllers/api/prosecution_cases_controller_spec.rb
archferns: The controller specs are only validating the response status codes, while the request specs do a bit thorough check of the schema. I would want to keep these in to easily ensure that we have non-happy status codes tested

spec/requests/api/prosecution_cases_spec.rb
archferns: Good idea,
I made a subtask to move this in https://dsdmoj.atlassian.net/browse/CACP-165 (via a new PR so as to avoid this one getting large)

archferns: Added via https://github.com/ministryofjustice/laa-court-data-adaptor/pull/63 
Thanks @mpw5

PR #66, opened 2020-02-10
CACP-166 API Versioning

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-166)

Adds the `versionist` gem to allow for versioning of API endpoints. This is implemented at present to use the URL path to define the version - but this can easily be changed to use an HTTP header or request parameter if required.

As a result the code has been refactored to move functionality into the `api` namespace. Some additional refactoring was done to introduce the idea of `internal` and `external` namespaces (ie those available to other LAA applications and HMCTS applications respectively). This allows for better control of endpoints offering similar functionality (eg. we will have two `hearings` endpoints - one to be called by CPUI for searching, the other to be called by HMCTS to post updates. This approach allows the two to be kept clearly separate.)

The URLS for the two existing API endpoints are now

`/api/internal/v1/prosecution_cases` (was `/api/prosecution_cases`)

and 

`/api/external/v1/hearings` (was `/hearings`)

This PR also removes an unnecessary route/controller and associated specs, and fixes an issue with `binding.pry`.

## Checklist


mpw5:
> Looks great 👍
> We should probably also make changes to the schema before we merge this

Good point. I forgot about the schema. I'll update it and add it to this PR.

PR #75, opened 2020-02-19
Add hearing summary to prosecution case

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-CACP88)

Hearing summaries are part of the search response returned by Common Platform. This PR adds a `HearingSummary` model and serializer, and includes it in the existing `prosecution_case` model.

## Checklist

### DIFF
--- a/app/models/defendant.rb
+++ b/app/models/defendant.rb
@@ -3,7 +3,7 @@
 class Defendant
   include ActiveModel::Model
 
-  attr_accessor :body
+  attr_accessor :body, :gender, :address_1, :address_2, :address_3, :address_4, :address_5, :postcode
 
   def id
     body['defendantId']
@@ -23,34 +23,6 @@ class Defendant
 
   def national_insurance_number
     body['nationalInsuranceNumber']
-  end
-
-  def gender
-    body['gender']
-  end
-
-  def address_1
-    body['address_1']
-  end
-
-  def address_2
-    body['address_2']
-  end
-
-  def address_3
-    body['address_3']
-  end
-
-  def address_4
-    body['address_4']
-  end
-
-  def address_5
-    body['address_5']
-  end
-
-  def postcode
-    body['postcode']
   end
 
   def offences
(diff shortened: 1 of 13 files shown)


app/models/hearing_summary.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+class HearingSummary
+  include ActiveModel::Model
+
+  attr_accessor :body, :hearing_date
```
archferns: I think we are missing where we pass `hearing_date` to the model as well

app/models/defendant.rb
```diff
@@ -3,7 +3,7 @@
 class Defendant
   include ActiveModel::Model
 
-  attr_accessor :body
+  attr_accessor :body, :gender, :address_1, :address_2, :address_3, :address_4, :address_5, :postcode
```
archferns: I think we are missing the initilization here (apart from the `body`)

mpw5: `:gender, :address_1, :address_2, :address_3, :address_4, :address_5, :postcode` aren't actually part of the defendant data we get from the prosecution_case search results, so they won't appear in the body. That's the data we need to get by doing subsequent calls to get additional hearing data. 

My original thought was that we could update the `body` with that data after doing those calls, but after thinking about it it seemed easier to hold them as separate attributes on the model, hence this change.

You're right that they don't get initialised here, they would only get set later on, after we've done those extra hearing calls.

app/models/hearing_summary.rb
mpw5: Similar to the `defendant` model, the `hearing_date` isn't data we receive in the `body` of a hearing summary, it's supplemental data from a subsequent hearing call, so I've added it as an additional attribute that can be populated when that data is available.

In both cases maybe I shouldn't have included them yet, as they're part of the adaptations which are still to be done. If it's cleaner I can remove them from both models for now and add them in a later PR.

archferns: That makes sense. I'm happy to have them in as it complements your serializer specs.

app/models/defendant.rb
archferns: makes sense. I'm happy to have them in as it complements your serializer specs.

PR #84, opened 2020-02-26
Expose offence info to serializer

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-88)

Added these attributes from offence:
`orderIndex`, `offenceCode`, `offenceTitle`, `modeOfTrial`

Pending: (These are available probably from the GET hearings call, so, perhaps we could spike and add them once we have figured out the absolute required data set)
`.plea.pleaDate`, `.plea.pleaValue`, `.count`

 Allow exposing offences as a relationship on defendants, hence we can make queries of the form:

`include=defendants,defendants.offences` to be able to output both defendant and defendant offences

## Checklist

### DIFF
--- a/app/models/prosecution_case.rb
+++ b/app/models/prosecution_case.rb
@@ -11,14 +11,6 @@ class ProsecutionCase < ApplicationRecord
     defendants.map(&:id)
   end
 
-  def hearing_summaries
-    body['hearings'].map { |hearing_summary| HearingSummary.new(body: hearing_summary) }
-  end
-
-  def hearing_summary_ids
-    hearing_summaries.map(&:id) if body['hearings'].present?
-  end
-
   def prosecution_case_reference
     body['prosecutionCaseReference']
   end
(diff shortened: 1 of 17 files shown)


app/models/prosecution_case.rb
```diff
@@ -11,14 +11,6 @@ def defendant_ids
...
   end
 
-  def hearing_summaries
-    body['hearings'].map { |hearing_summary| HearingSummary.new(body: hearing_summary) }
-  end
-
-  def hearing_summary_ids
-    hearing_summaries.map(&:id) if body['hearings'].present?
-  end
-
```
mpw5: Just a question, not a problem, I was just wondering why we're removing the link between `prosecution_cases` and `hearing_summaries`? Is there a better way of modelling them?

archferns: I removed it because I was going from the Miro board, and didn't find `hearings` listed anywhere in the current design, but maybe I haven't fully understood how they are represented in the UI.
I will check again to be sure.

archferns: Closing in favour of handling the post linking search later.

PR #89, opened 2020-03-04
Add POST implementation for recording an LAA reference

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-230)
Add a POST `#create` endpoint for the UI to be able to record an LAA reference against a defendant.
Dependency additions:
1. [dry-validation](https://github.com/dry-rb/dry-validation): to validate input 
2. [jsonapi_parameters)](https://github.com/visualitypl/jsonapi_parameters), which is a micro deserilaizer.
The create endpoint will receive the input from the UI, validate it, and on a sucessful validation, return a `202 Accepted` status response.
In case of invalid input, it will return with a `400 Bad Request`

## Checklist


spec/services/laa_reference_creator_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  it 'validates the payload' do
+    expect(subject.errors).to be_blank
+  end
+
+  context 'with a non-numeric maat_reference' do
+    let(:maat_reference) { 'ABC123123' }
+
+    it 'raises errors against the payload' do
+      expect(subject).not_to be_blank
```
mpw5: Should this be `subject.errors`?

archferns: Yes, that is indeed correct, good spot, thanks

jsugarman: `expect(subject)` is the same as `is_expected`, but whatever is your preference.

archferns:
Thanks, I went with dry-validation as it is meant to be a [suitable replacement for both strong-params and validations](https://dry-rb.org/gems/dry-schema/1.4/#when-to-use).
Apart from that, we are still spiking the actual call to CP/MLRA, so this might change to use something more favourable/standard if we have to in the long run.

PR #90, opened 2020-03-05
CACP-231 Record prosecution_case/defendant/offence IDs

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-231)

First part of CACP-231. The overall goal of this story is to map a single link request from View Court Data into multiple calls to Common Platform. As a first step this PR creates a model and database table to map `prosecution_case_ids` to `defendant_ids` to `offence_ids`, and adds a call to the `ProsecutionCaseRecorder` to create records. This will enable us to identify which offences need to be included in the link calls to Common Platform.

## Checklist

### DIFF
--- a/app/services/prosecution_case_recorder.rb
+++ b/app/services/prosecution_case_recorder.rb
@@ -9,6 +9,13 @@ class ProsecutionCaseRecorder < ApplicationService
 
   def call
     prosecution_case.update(body: body)
+
+    prosecution_case.defendants.each do |defendant|
+      defendant.offences.each do |offence|
+        ProsecutionCaseDefendantOffence.find_or_create_by!(prosecution_case_id: prosecution_case.id, defendant_id: defendant.id, offence_id: offence.id)
+      end
+    end
+
     prosecution_case
   end
 
--- a/db/migrate/20200304144205_create_prosecution_case_defendant_offences.rb
+++ b/db/migrate/20200304144205_create_prosecution_case_defendant_offences.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateProsecutionCaseDefendantOffences < ActiveRecord::Migration[6.0]
+  def change
+    create_table :prosecution_case_defendant_offences, id: :uuid do |t|
+      t.uuid :prosecution_case_id,    null: false
+      t.uuid :defendant_id,           null: false
+      t.uuid :offence_id,             null: false
+
+      t.timestamps
+    end
+  end
+end
--- a/app/models/prosecution_case_defendant_offence.rb
+++ b/app/models/prosecution_case_defendant_offence.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class ProsecutionCaseDefendantOffence < ApplicationRecord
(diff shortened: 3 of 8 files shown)


app/services/prosecution_case_recorder.rb
```diff
@@ -9,6 +9,13 @@ def initialize(prosecution_case_id, body)
 
   def call
     prosecution_case.update(body: body)
```
archferns: I'm wondering if this can be taken off now?

mpw5: If I just remove it a load of tests fail. We're using the data in the `body` in a few places - for example in the `ProsecutionCase` model:  
```
def defendants
  body['defendants']&.map { |defendant| Defendant.new(body: defendant) }
end
```

archferns: oh, yes, then not a good idea :+1:

db/migrate/20200304144205_create_prosecution_case_defendant_offences.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateProsecutionCaseDefendantOffences < ActiveRecord::Migration[6.0]
+  def change
+    create_table :prosecution_case_defendant_offences, id: :uuid do |t|
+      t.string :prosecution_case_id,    null: false
+      t.string :defendant_id,           null: false
```
archferns: Should we have these as `uuid`s instead?
also, perhaps one for later when we loop: index for `defendant_id`

app/models/prosecution_case.rb
```diff
@@ -4,7 +4,7 @@ class ProsecutionCase < ApplicationRecord
   validates :body, presence: true
 
   def defendants
-    body['defendants'].map { |defendant| Defendant.new(body: defendant) }
+    body['defendants']&.map { |defendant| Defendant.new(body: defendant) }
```
archferns: I'm wondering if we need a safe navigation operator here, considering that the [defendants are available in the schema](https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/6955ae92a00efe5adc924eba2f5aaf3e658b97f9/lib/schemas/global/search/prosecutionCaseSummary.json#L39)

app/models/defendant.rb
```diff
@@ -30,7 +30,7 @@ def arrest_summons_number
   end
 
   def offences
-    body['offences'].map { |offence| Offence.new(body: offence) }
+    body['offences']&.map { |offence| Offence.new(body: offence) }
```
archferns: Similar to `defendants` shouldn't we assume that we have this info anyway?

db/migrate/20200304144205_create_prosecution_case_defendant_offences.rb
mpw5: Changed to UUIDs

app/models/prosecution_case.rb
mpw5: Removed

app/models/defendant.rb
mpw5: Removed

PR #91, opened 2020-03-06
Make CP Link calls on request from View Court Data

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-231)

View Court Data provides a `maat_reference` and `defendant_id` in its link requests. CDA needs to translate that `defendant_id` into the (potentially) multiple `offence_id`s that belong to that defendant and make a call to common platform for each one.

This PR amends the `laa_references_controller` so that when a link request is received from View Court Data, the correct `offence_id`s for that defendant are identified and a call is made to Common Platform to register the `maat_reference` against each one.

## Checklist

### DIFF
--- a/app/controllers/api/internal/v1/laa_references_controller.rb
+++ b/app/controllers/api/internal/v1/laa_references_controller.rb
@@ -9,8 +9,7 @@ module Api
           if contract.errors.present?
             render json: contract.errors.to_hash, status: :bad_request
           else
-            make_common_platform_calls(contract)
-
+            LaaReferenceUpdaterJob.perform_later(contract)
             render status: :accepted
           end
         end
@@ -27,25 +26,6 @@ module Api
 
         def transformed_params
           create_params.slice(*allowed_params).to_hash.transform_keys(&:to_sym)
-        end
-
-        def make_common_platform_calls(contract)
-          maat_reference = contract[:maat_reference]
-          defendant_id = contract[:defendant_id]
-          offences = ProsecutionCaseDefendantOffence.where(defendant_id: defendant_id)
-          prosecution_case_id = offences.pluck(:prosecution_case_id).first
-          offence_ids = offences.pluck(:offence_id)
-
-          offence_ids.each do |offence_id|
-            LaaReferenceUpdaterJob.perform_later(
-              prosecution_case_id: prosecution_case_id,
-              defendant_id: defendant_id,
-              offence_id: offence_id,
-              status_code: 'AP',
-              application_reference: maat_reference,
-              status_date: Date.today.strftime('%Y-%m-%d')
-            )
-          end
         end
       end
     end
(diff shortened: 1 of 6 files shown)


app/controllers/api/internal/v1/laa_references_controller.rb
```diff
@@ -26,6 +28,25 @@ def allowed_params
         def transformed_params
           create_params.slice(*allowed_params).to_hash.transform_keys(&:to_sym)
         end
+
+        def make_common_platform_calls(contract)
+          maat_reference = contract[:maat_reference]
+          defendant_id = contract[:defendant_id]
+          offences = ProsecutionCaseDefendantOffence.where(defendant_id: defendant_id)
+          prosecution_case_id = offences.pluck(:prosecution_case_id).first
```
archferns: since `ProsecutionCaseDefendantOffence` contain all the required info, perhaps we could loop through the `offences` itself , as `pluck` might be making extra db calls.
eg: 
```
ProsecutionCaseDefendantOffence.where(defendant_id: defendant_id).each do |case|
    LaaReferenceUpdaterJob.call(case.offence_id, case.prosecution_case_id, ....
    ...
end
```

app/controllers/api/internal/v1/laa_references_controller.rb
```diff
@@ -9,6 +9,8 @@ def create
           if contract.errors.present?
             render json: contract.errors.to_hash, status: :bad_request
           else
+            make_common_platform_calls(contract)
```
archferns: I believe this can be extracted out to a service that we can then call as a job:
```
LaaReferenceUpdaterJob.perform(contract)
```

app/jobs/laa_reference_updater_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class LaaReferenceUpdaterJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    Api::RecordLaaReference.new(*args).call
```
archferns: perhaps
```
MakeCommonPlatformCalls.call...
```

app/controllers/api/internal/v1/laa_references_controller.rb
mpw5: Yes, that's neater. Thanks.

app/jobs/laa_reference_updater_job.rb
mpw5: Do you mean add a `MakeCommonPlatformCalls` service class, and call that from the job?

So I could move the logic from the `make_common_platform_calls` method in `LaaReferencesController` to that new service class which would do the looping and make the requests, and call that from the `LaaReferenceUpdaterJob`, which would be called from `LaaReferencesController`.

I think I was considering it slightly differently, where each request to CP would be a new job - so each request is made asynchronously rather than having a single async job making a number of requests.

So I alternatively I could move the logic from the `make_common_platform_calls` method in `LaaReferencesController` to a new service, which would be called from `LaaReferencesController` and would loop through the offences, but would create a new `LaaReferenceUpdaterJob` for each offence to actually make the call to CP.

But I'm not sure I have a strong opinion on it either way, if you can see a drawback to my approach I'm happy to go with what you suggest.

PR #92, opened 2020-03-10
CACP-184 Link with no MAAT reference

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-184)

Add a sequence to the database. If a request is made to link a case without a MAAT reference, make a call to the database to get the next sequence value, format it correctly and use it as the MAAT reference.

## Checklist


spec/services/laa_reference_updater_spec.rb
```diff
@@ -33,4 +36,28 @@
       record
     end
   end
+
+  context 'with no maat reference' do
+    let(:maat_reference) { nil }
+
+    before do
+      allow_any_instance_of(LaaReferenceUpdater).to receive(:dummy_maat_reference).and_return('A1000000')
```
archferns:
```suggestion
      ActiveRecord::Base.connection.execute('ALTER SEQUENCE dummy_maat_reference_seq RESTART;')
```
Would it be easier if we reset the sequence, and then we can drop the private method spec.

spec/services/laa_reference_updater_spec.rb
```diff
@@ -6,7 +6,10 @@
   let(:contract) { NewLaaReferenceContract.new.call({ maat_reference: maat_reference, defendant_id: defendant_id }) }
   let(:mock_record_laa_reference_service) { double Api::RecordLaaReference }
```
archferns: perhaps this can be an [`instance_double`](https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/verifying-doubles/using-an-instance-double)?
```suggestion
  let(:mock_record_laa_reference_service) { instance_double Api::RecordLaaReference }
```

db/structure.sql
```diff
@@ -0,0 +1,325 @@
...
+    ADD CONSTRAINT fk_rails_b4b53e07b8 FOREIGN KEY (application_id) REFERENCES public.oauth_applications(id);
+
+
+--
+-- PostgreSQL database dump complete
+--
+
+SET search_path TO "$user", public;
+
+INSERT INTO "schema_migrations" (version) VALUES
```
archferns: This is amazing!
I think this is much more clearer than the standard schema.rb.

spec/services/laa_reference_updater_spec.rb
mpw5: Yes, this is much better

spec/services/laa_reference_updater_spec.rb
mpw5: Fixed!

db/structure.sql
mpw5: 🎉

PR #93, opened 2020-03-11
Extend prosecution_case_defendant_offences table

mpw5:
## What

See comments on [CACP-89](https://dsdmoj.atlassian.net/browse/CACP-89) and [CACP-184](https://dsdmoj.atlassian.net/browse/CACP-184)

We want to capture additional data in the `prosecution_case_defendant_offences` table to provide more of an audit trail. 

Based on Richard's comments on the above ticket I've added the following fields to the table:
 - `maat_reference`: the maat id
 - `dummy_maat_reference`: boolean to indicate if a dummy maat id is used
 - `rep_order_status`: the legal aid status for the offence
 - `response_status`: HTTP status from the common platform response
 - `response_body`: json representation of the common platform response body

## Checklist

### DIFF
--- a/app/services/laa_reference_updater.rb
+++ b/app/services/laa_reference_updater.rb
@@ -6,34 +6,26 @@ class LaaReferenceUpdater < ApplicationService
   end
 
   def call
-    @maat_reference = contract[:maat_reference] || dummy_maat_reference
-
     ProsecutionCaseDefendantOffence.where(defendant_id: contract[:defendant_id]).each do |offence|
-      @response = Api::RecordLaaReference.call(
+      Api::RecordLaaReference.call(
         prosecution_case_id: offence.prosecution_case_id,
         defendant_id: offence.defendant_id,
         offence_id: offence.offence_id,
         status_code: 'AP',
-        application_reference: @maat_reference,
+        application_reference: maat_reference,
         status_date: Date.today.strftime('%Y-%m-%d')
       )
-
-      update_record(offence)
     end
   end
 
   private
 
-  def dummy_maat_reference
-    "A#{ActiveRecord::Base.connection.execute("SELECT nextval('dummy_maat_reference_seq')")[0]['nextval']}"
+  def maat_reference
+    @maat_reference ||= contract[:maat_reference].presence || dummy_maat_reference
   end
 
-  def update_record(offence)
-    offence.maat_reference = @maat_reference
-    offence.dummy_maat_reference = (@maat_reference.to_s[0] == 'A')
-    offence.response_status = @response.status
-    offence.response_body = @response.body
-    offence.save!
+  def dummy_maat_reference
+    "A#{ActiveRecord::Base.connection.execute("SELECT nextval('dummy_maat_reference_seq')")[0]['nextval']}"
   end
 
   attr_reader :contract
--- a/app/services/laa_reference_updater.rb
+++ b/app/services/laa_reference_updater.rb
@@ -6,17 +6,19 @@ class LaaReferenceUpdater < ApplicationService
   end
 
   def call
-    maat_reference = contract[:maat_reference]
+    @maat_reference = contract[:maat_reference] || dummy_maat_reference
 
     ProsecutionCaseDefendantOffence.where(defendant_id: contract[:defendant_id]).each do |offence|
-      Api::RecordLaaReference.call(
+      @response = Api::RecordLaaReference.call(
         prosecution_case_id: offence.prosecution_case_id,
         defendant_id: offence.defendant_id,
         offence_id: offence.offence_id,
         status_code: 'AP',
-        application_reference: maat_reference || dummy_maat_reference,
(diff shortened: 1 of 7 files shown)


db/migrate/20200310222258_add_columns_to_prosecution_case_defendant_offences.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddColumnsToProsecutionCaseDefendantOffences < ActiveRecord::Migration[6.0]
+  def change
+    add_column :prosecution_case_defendant_offences, :maat_reference, :string
+    add_column :prosecution_case_defendant_offences, :dummy_maat_reference, :boolean
```
archferns:
```suggestion
    add_column :prosecution_case_defendant_offences, :dummy_maat_reference, :boolean, default: false, null: false
```

db/migrate/20200310222258_add_columns_to_prosecution_case_defendant_offences.rb
```diff
@@ -0,0 +1,11 @@
...
+    add_column :prosecution_case_defendant_offences, :response_status, :integer
+    add_column :prosecution_case_defendant_offences, :response_body, :json
+    add_column :prosecution_case_defendant_offences, :user_id, :string
```
archferns: This is an interesting one, at the moment, we authenticate with the UI via `client_credentials`, but I guess we can leave it to the UI to inform us of the requesting user.
In terms of the user format, should we probably agree with the UI on our their format before we add this one?

app/services/laa_reference_updater.rb
```diff
@@ -6,17 +6,19 @@ def initialize(contract)
...
         prosecution_case_id: offence.prosecution_case_id,
         defendant_id: offence.defendant_id,
         offence_id: offence.offence_id,
         status_code: 'AP',
-        application_reference: maat_reference || dummy_maat_reference,
+        application_reference: @maat_reference,
         status_date: Date.today.strftime('%Y-%m-%d')
       )
+
+      update_record(offence)
```
archferns: Could this be better suited to be triggered from within the `Api::RecordLaaReference` call?
similar to [the get hearing call](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/1777cd4b39f9baeac8d28d4e157cb98c21891956/app/services/api/get_hearing_results.rb)

app/services/laa_reference_updater.rb
```diff
@@ -6,17 +6,19 @@ def initialize(contract)
   end
 
   def call
-    maat_reference = contract[:maat_reference]
+    @maat_reference = contract[:maat_reference] || dummy_maat_reference
```
archferns:
```suggestion
    maat_reference = contract[:maat_reference].presence || dummy_maat_reference
```
I think we should make this private as well

db/migrate/20200310222258_add_columns_to_prosecution_case_defendant_offences.rb
mpw5: 👍

db/migrate/20200310222258_add_columns_to_prosecution_case_defendant_offences.rb
mpw5: Fair point. I'll remove it for now and add a ticket to add it in later when we've agreed how this will work

app/services/laa_reference_updater.rb
mpw5: 👍

app/services/laa_reference_updater.rb
mpw5: Moved to `Api::RecordLaaReference`

mpw5:
> 1. I think we should keep the `user_id` out for now. Probably better if we dictate the format to the UI (eg: uuid), because later on we probably also need to do something similar for the maat api sending data over to us, and that will make us guarentee uniqueness.

Good point. It felt a bit premature to be including it so I'm happy to remove it.

> 2. For `Any previous LAA Reference (MAAT ID)/Rep Order Status`, we would need to have an audit trail of previously linked references. Perhaps we need a ticket to cover that separately?

Yes, this was one of Richard's suggestions that I wanted to defer. It felt potentially quite complex, and I thought it might make more sense to address it after we've implemented unlinking. I'll create a new ticket for it.

app/services/laa_reference_updater.rb
```diff
@@ -6,22 +6,24 @@ def initialize(contract)
   end
 
   def call
-    maat_reference = contract[:maat_reference]
-
     ProsecutionCaseDefendantOffence.where(defendant_id: contract[:defendant_id]).each do |offence|
-      Api::RecordLaaReference.call(
+      @response = Api::RecordLaaReference.call(
```
archferns:
```suggestion
      Api::RecordLaaReference.call(
```

PR #95, opened 2020-03-12
Add an SQS publisher

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-232)

WIP for CACP-232.
Pending:
1. Add the actual JSON contents to be sent to the queue.
2. Add relevant environment variables in the helm values.

## Checklist


spec/services/sqs/publish_laa_reference_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+          }
+        ]
+      },
+      sessions: []
+    }
+  end
+
+  subject { described_class.call(prosecution_case_id: prosecution_case_id, defendant_id: defendant_id, maat_reference: maat_reference) }
+
+  it 'triggers apublish call with the sqs payload' do
```
mpw5: Really nitpicky - i guess `apublish` should be `a publish`

spec/services/sqs/publish_laa_reference_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    expect(Sqs::MessagePublisher).to receive(:call).with(message: sqs_payload)
+    subject
+  end
+
+  context 'with a dummy maat_reference' do
+    it 'triggers apublish call with the sqs payload' do
```
mpw5: As above

spec/services/sqs/publish_laa_reference_spec.rb
archferns: Good catch, updated!

spec/services/sqs/publish_laa_reference_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    ProsecutionCase.create!(
+      id: prosecution_case_id,
+      body: JSON.parse(file_fixture('prosecution_case_search_result.json').read)
+    )
+  end
+  let(:maat_reference) { 123_456 }
+
+  let(:sqs_payload) do
+    {
+      maatId: 123_456,
```
mpw5: Could this use the `maat_reference` variable defined above? (Or remove that `let` block because it doesn't seem to be being used anywhere else)

spec/services/sqs/publish_laa_reference_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    expect(Sqs::MessagePublisher).to receive(:call).with(message: sqs_payload)
+    subject
+  end
+
+  context 'with a dummy maat_reference' do
+    it 'triggers a publish call with the sqs payload' do
+      expect(Sqs::MessagePublisher).to receive(:call).with(message: sqs_payload)
+      subject
+    end
+  end
```
mpw5: I might be missing something but t looks like this test does exactly the same as the one above it. Do we need something like `let(:maat_reference) { 'A123456' }` in there to test that it actually works with a dummy id?

spec/services/sqs/publish_laa_reference_spec.rb
archferns: Updated to use `maat_reference`

spec/services/sqs/publish_laa_reference_spec.rb
archferns: Good spot, this was a duplicate spec, which isn't relevant anymore

PR #106, opened 2020-03-24
Add dev config for AWS SQS

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-LASB232)

Add config for Dev environment, and add temporary values for missing/unmapped mandatory MLRA attributes.

## Checklist

### DIFF
--- a/app/services/sqs/publish_laa_reference.rb
+++ b/app/services/sqs/publish_laa_reference.rb
@@ -2,6 +2,15 @@
 
 module Sqs
   class PublishLaaReference < ApplicationService
+    TEMPORARY_CJS_AREA_CODE = 16
+    TEMPORARY_CREATED_USER = 'cpUser'
+    TEMPORARY_CJS_LOCATION = 'B16BG'
+    TEMPORARY_DATE_OF_HEARING = '2020-08-16'
+    TEMPORARY_POST_HEARING_CUSTODY = 'R'
+    TEMPORARY_OFFENCE_CODE = 'AA06035'
+    TEMPORARY_MODE_OF_TRIAL = 1
+    TEMPORARY_RESULT_CODE = 3026
+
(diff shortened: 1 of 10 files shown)


app/services/sqs/publish_laa_reference.rb
```diff
@@ -2,6 +2,15 @@
 
 module Sqs
   class PublishLaaReference < ApplicationService
+    TEMPORARY_CJS_AREA_CODE = 16
+    TEMPORARY_CJS_USER = 'cpUser'
+    TEMPORARY_CJS_LOCATION = 'B16BG'
```
mpw5: I think maybe `TEMPORARY_USER` might be a better name here. I _think_ that the user details we're interested here are internal LAA users, so wouldn't really fall under CJS. No big deal but might just avoid confusion.

helm_deploy/laa-court-data-adaptor/values-dev.yaml
```diff
@@ -14,6 +14,7 @@ nameOverride: ""
 fullnameOverride: ""
 
 common_platform_url: https://hmcts-common-platform-mock-api.apps.live-1.cloud-platform.service.justice.gov.uk
+sqs_region: eu-west-2
```
mpw5: Is it worth adding this to the `values-test`, `values-uat`, `values-stage` and `values-prod` files now too, just so they don't get forgotten?

app/services/sqs/publish_laa_reference.rb
archferns: Good point, It is definitely clearer.

helm_deploy/laa-court-data-adaptor/values-dev.yaml
archferns: Added

PR #107, opened 2020-03-24
CACP-235 Push unlink requests to SQS

mpw5:
#  What

[CACP-235](https://dsdmoj.atlassian.net/browse/CACP-235)

Create a `PublishUnlinkLaaReference` class which takes a `maat_reference`, `user_id`, `unlink_reason_id`, and `unlink_reason_text` and creates a message which is pushed to the SQS queue for MAAT-API to consume.

This will be called from a service class which doesn't currently exist yet, so hasn't been wired up. That can be done as part of [CACP-234](https://dsdmoj.atlassian.net/browse/CACP-234) or as a separate change later. 

## Checklist

### DIFF
--- a/app/services/sqs/publish_unlink_laa_reference.rb
+++ b/app/services/sqs/publish_unlink_laa_reference.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Sqs
+  class PublishUnlinkLaaReference < ApplicationService
+    def initialize(maat_reference:, user_id:, unlink_reason_id:, unlink_reason_text:)
+      @maat_reference = maat_reference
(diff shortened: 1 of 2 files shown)


archferns:
It is in the correct direction, one change we need to make is to map it to the correct queue.
MLRA/MAAT API uses separate queues for each of the tasks.
We might want to configure the Message publisher to identify which queue to use accordingly:
https://dsdmoj.atlassian.net/wiki/spaces/LAACP/pages/1755939335/SQS+Crime-Apps+queues

mpw5:
Superseded by https://github.com/ministryofjustice/laa-court-data-adaptor/pull/114

Issue #113, opened 2020-03-26
Expose maat_reference on defendant


jsugarman:
**What**
expose the `maat_reference` attribute on the `defendant`

**Why**
for simple display in the UI

**Also**
This could replace the `is_linked` attribute on defendant as `nil` maat_reference would mean
is not linked and vice-versa.

PR #115, opened 2020-03-27
Expose maat_reference on defendant

mpw5:
## What

Fixes https://github.com/ministryofjustice/laa-court-data-adaptor/issues/113

Replace the `is_linked` attribute on the defendant model with `maat_reference`. This returns the `maat_reference` of the defendant's first offence unless the `maat_reference` begins with 'Z' - this indicates that the case has been unlinked, and so `nil` is returned instead.

`is_linked` is no longer required as View Court Data can infer this from the presence (or not) of a `maat_reference`.

## Checklist

### DIFF
--- a/app/models/defendant.rb
+++ b/app/models/defendant.rb
@@ -37,7 +37,9 @@ class Defendant
     offences.map(&:id)
   end
 
-  def linked?
-    offences.map(&:maat_reference).compact.present?
+  def maat_reference
+    return nil if offences == [] || offences.first.maat_reference&.first == 'Z'
+
+    offences.first.maat_reference
   end
 end
--- a/app/models/defendant.rb
+++ b/app/models/defendant.rb
@@ -38,8 +38,15 @@ class Defendant
   end
 
   def maat_reference
-    return nil if offences == [] || offences.first.maat_reference&.first == 'Z'
+    _maat_reference unless _maat_reference&.start_with?('Z')
+  end
 
-    offences.first.maat_reference
+  private
+
+  def _maat_reference
+    refs = offences.map(&:maat_reference).uniq.compact
+    raise 'Too many maat references' if refs.size > 1
+
+    refs&.first
   end
 end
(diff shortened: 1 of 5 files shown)


mpw5:
Thanks @archferns. Good point. I've updated `swagger/v1/defendant.json` too.

app/models/defendant.rb
```diff
@@ -37,7 +37,9 @@ def offence_ids
     offences.map(&:id)
   end
 
-  def linked?
-    offences.map(&:maat_reference).compact.present?
+  def maat_reference
+    return nil if offences == [] || offences.first.maat_reference&.first == 'Z'
```
jsugarman:
```suggestion
    return nil if offences.empty? || offences.first.maat_reference&.first&.eql?( 'Z')
```

jsugarman: is it ever possible to have different `maat_references` on offences?

perhaps there should be something like below. and if it ever returns more that one then an error should be raise
```
offences.map(&:maat_reference).uniq.compact
```

jsugarman: **^^ not a blocker**

spec/models/defendant_spec.rb
```diff
@@ -21,7 +21,20 @@
...
+    let(:offences) do
+      [instance_double('Offence', maat_reference: nil),
+       instance_double('Offence', maat_reference: nil)]
+    end
+
+    before do
+      allow(defendant).to receive(:offences).and_return(offences)
+    end
+
+    it { expect(defendant.maat_reference).to eq(nil) }
```
jsugarman:
```suggestion
    it { expect(defendant.maat_reference).to be_nil }
```
https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/be-matchers

app/models/defendant.rb
jsugarman: just to mention that `string.first` is a [rails extension](https://api.rubyonrails.org/classes/String.html#method-i-first).
consider alternative `start_with?` ruby string class method which may be clearer:
```
offences.first.maat_reference.start_with?('Z')
```

jsugarman: have added a commit to handle these @mpw5

PR #119, opened 2020-04-02
Add a RepresentationOrder endpoint with validations

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-LASB185)

1. Added a representation order endpoint with validations to ensure the dataset sent across matches the CP validations.
2. To ensure validations match CP, I am using `json-schema` (which was previously used only in testing), so that it can be matched directly against the CP schema constraints.
3. Added swagger docs for this endpoint.

## Checklist


app/contracts/new_representation_order_contract.rb
```diff
@@ -0,0 +1,74 @@
...
+  rule(:maat_reference) do
+    key.failure('must not exceed 9 digits') if key? && value > 999_999_999
+  end
+
+  rule('defence_organisation.organisation.address.postcode') do
+    key.failure('is not a valid postcode') if key? && !postcode_validator.call(postcode: value)
+  end
+
+  rule('defence_organisation.organisation.contact.primary_email') do
+    key.failure('is not a valid email') if key? && !email_validator.call(email: value)
```
jsugarman, line 56: **not a blocker**
```suggestion
    key.failure('is not a valid email') if [key?, !email_validator.call(email: value)].all?
```
_not convinced this is clearer/simpler but just putting it out there!_

app/services/api/record_representation_order.rb
```diff
@@ -42,10 +44,11 @@ def request_body
         applicationReference: application_reference,
         statusDate: status_date,
         effectiveStartDate: effective_start_date,
+        effectiveEndDate: effective_end_date,
         defenceOrganisation: defence_organisation
-      }
+      }.compact
```
jsugarman, line 49: just a question: is for request to CP and isis this a general approach you are adopting - to not send a key if it is `nil` value?

app/services/common_platform/email_validator.rb
```diff
@@ -0,0 +1,24 @@
...
+    def call
+      JSON::Validator.validate(schema, email, validate_schema: true)
+    end
+
+    private
+
+    attr_reader :email
+
+    def schema
+      {
```
jsugarman, line 18: **not a blocker** just a a suggestion to consider: is there any way to tie together the swagger json and the schema snippet here (to use snippet in swagger) - so that there not two sources for these regex patterns

app/services/api/record_representation_order.rb
archferns: Yep, the general approach is to not send the key for `nil` values, since the presence of optional values expects them to validate against the schema. 
More specifically, in this case `effective_end_date` sent across to HMCTS is optional, but if present, should be a date. https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/0bd1cb06934abe89012c477fcb67c9167a0e1c91/lib/schemas/api/progression.recordRepresentationOrder.json#L35-L38

app/services/common_platform/phone_validator.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    private
+
+    attr_reader :phone
+
+    def schema
+      {
+        "type": 'string',
+        "pattern": '^[\\+]?[0-9()\\-\\.\\s]+$'
+      }
```
jsugarman, line 21: as [above](https://github.com/ministryofjustice/laa-court-data-adaptor/pull/119/files#r403102070)

app/services/common_platform/postcode_validator.rb
```diff
@@ -0,0 +1,28 @@
...
+    attr_reader :postcode
+
+    # rubocop:disable Layout/LineLength
+    def schema
+      {
+        "description": 'UK Gov post code validation following CJS Data Standards',
+        "type": 'string',
+        "pattern": '^(([gG][iI][rR] {0,}0[aA]{2})|(([aA][sS][cC][nN]|[sS][tT][hH][lL]|[tT][dD][cC][uU]|[bB][bB][nN][dD]|[bB][iI][qQ][qQ]|[fF][iI][qQ][qQ]|[pP][cC][rR][nN]|[sS][iI][qQ][qQ]|[iT][kK][cC][aA]) {0,}1[zZ]{2})|((([a-pr-uwyzA-PR-UWYZ][a-hk-yxA-HK-XY]?[0-9][0-9]?)|(([a-pr-uwyzA-PR-UWYZ][0-9][a-hjkstuwA-HJKSTUW])|([a-pr-uwyzA-PR-UWYZ][a-hk-yA-HK-Y][0-9][abehmnprv-yABEHMNPRV-Y]))) [0-9][abd-hjlnp-uw-zABD-HJLNP-UW-Z]{2}))$',
+        "maxLength": 8
+      }
```
jsugarman, line 24: as [above](https://github.com/ministryofjustice/laa-court-data-adaptor/pull/119/files#r403102070)

app/services/common_platform/email_validator.rb
archferns: I have been considering adding the HMCTS schemas itself as maybe a gem/some sort of dependency between both the mock and CDA.
Using the schema snippet as well makes a lot of sense. I will make an investigation around this.

archferns: Added a [ticket](https://dsdmoj.atlassian.net/browse/CACP-308)

app/contracts/new_representation_order_contract.rb
archferns: Added this, seems cleaner :+1:

.rubocop.yml
```diff
@@ -26,6 +26,7 @@ Metrics/AbcSize:
 Metrics/BlockLength:
   Exclude:
   - 'config/routes.rb'
+  - 'app/contracts/new_representation_order_contract.rb'
```
richrace, line 29: Is there a specific reason for this file to be excluded?

app/services/api/record_representation_order.rb
```diff
@@ -42,10 +47,23 @@ def request_body
...
     end
 
-    attr_reader :url, :status_code, :application_reference, :status_date, :effective_start_date, :defence_organisation, :headers, :connection
+    def update_database(response)
+      offence = ProsecutionCaseDefendantOffence.find_by(offence_id: offence_id)
```
richrace, line 56: In an ideal world, `offence` would always be present. But the `#find_by` can return `nil`. Do we want to handle this?

spec/jobs/representation_order_creator_job_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      defence_organisation: 'HASH'
+    }
+  end
+  describe '#perform_later' do
+    subject(:job) do
+      described_class.perform_later(argument_hash)
+    end
+
+    it 'queues a call to update the laa reference' do
+      ActiveJob::Base.queue_adapter = :test
```
richrace, line 21: I would expect this to be in the `config/environments/test.rb` file

app/services/api/record_representation_order.rb
archferns: Good point, I've made a [follow up ticket](https://dsdmoj.atlassian.net/browse/CACP-313) for this as this calls for a bit of refactoring

.rubocop.yml
archferns: Added due to the size of the Rep order payload.
However, I'm happy to move it to the inline syntax (`# rubocop:disable/enable`) if that's what you mean.

spec/jobs/representation_order_creator_job_spec.rb
archferns: Moved to the test config

.rubocop.yml
richrace: Talked via Slack about this. Outcome: I missed the specific Cop that we're ignoring.

spec/jobs/representation_order_creator_job_spec.rb
richrace: 0bd65b5 resolves this.

PR #134, opened 2020-04-28
Correct folder structure for SimpleCov reporting

richrace:
## What

Update config for SimpleCov. Use Rails presets + add folders.

## Checklist

### DIFF
--- a/app/channels/application_cable/channel.rb
+++ b/app/channels/application_cable/channel.rb
@@ -1,6 +1,0 @@
-# frozen_string_literal: true
-
-module ApplicationCable
-  class Channel < ActionCable::Channel::Base
-  end
-end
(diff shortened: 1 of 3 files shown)


richrace:
Yup these two files are now being included and tested. 

Are we using them? If not, can we remove them?


Screenshot 2020-04-29 at 09 21 42

mpw5:
I'm pretty sure we're not using them, but I'm not sure what the implication of removing them would be (does Rails do anything with ActionCable under the hood?). I guess deleting the files and running the tests will tell us!

Or we could just exclude them from code coverage - `add_filter 'app/channels'` in the `SimpleCov.start` block.

archferns:
We can delete those, as this is mainly an API only app.

richrace:
Cool. I'll make a separate PR for those changes then rebase this branch.

PR #171, opened 2020-05-27
Relax validation on representation order data

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-369)

MAAT API may send invalid representation order data to CDA. At present this causes contract validation failures and rejection of the data, however it is difficult to resolve this in MAAT-API/MLRA.

Instead, this PR relaxes the contract validation and instead strips any invalid data from the payload before sending it to CP.

## Checklist

### DIFF
--- a/app/services/representation_order_creator.rb
+++ b/app/services/representation_order_creator.rb
@@ -6,6 +6,7 @@ class RepresentationOrderCreator < ApplicationService
     @maat_reference = maat_reference
     @defendant_id = defendant_id
     @defence_organisation = defence_organisation.deep_transform_keys { |key| key.to_s.camelize(:lower).to_sym }
+    sanitise_defence_organisation
   end
 
   def call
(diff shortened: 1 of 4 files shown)


app/services/representation_order_creator.rb
```diff
@@ -27,10 +27,42 @@ def call_cp_endpoint
         status_date: offence[:status_date],
         effective_start_date: offence[:effective_start_date],
         effective_end_date: offence[:effective_end_date],
-        defence_organisation: defence_organisation
+        defence_organisation: sanitised_defence_organisation
```
archferns, line 30: It could be worth sanitising this in the initialize call so as to have a single memoised instance

spec/services/representation_order_creator_spec.rb
```diff
@@ -94,4 +94,46 @@
...
+      }
+    end
+    let(:transformed_defence_organisation) do
+      {
+        organisation: {
+          address: {
+            postcode: nil
+          },
+          contact: {
+            home: nil,
```
archferns, line 123: I think CP will not accept nil/null, and we might need to remove them before sending across.

mpw5: I'm now calling `delete` on any elements that fail validation, rather than setting them to `nil`.

app/services/representation_order_creator.rb
mpw5: Moved to `initialize`

app/services/representation_order_creator.rb
```diff
@@ -32,5 +33,35 @@ def call_cp_endpoint
...
+  end
+
+  def sanitise_email_addresses
+    %i[primaryEmail secondaryEmail].each do |email|
+      defence_organisation[:organisation][:contact].delete(email) unless CommonPlatform::EmailValidator.new(email: contact.dig(email)).call
+    end
+  end
+
+  def sanitise_phone_numbers
+    %i[home work mobile fax].each do |number|
```
archferns, line 61:
```suggestion
    %i[home mobile].each do |number|
```
I believe this should suffice, as the [work](https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/f0b400488aa612c9ecae2ba440bcfd5514c47974/lib/schemas/global/apiContactNumber.json#L11-L13)/[fax](https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/f0b400488aa612c9ecae2ba440bcfd5514c47974/lib/schemas/global/apiContactNumber.json#L27-L30) numbers can be strings

mpw5: 👍

PR #181, opened 2020-06-08
CACP334 Crown Court Notifications - Trials

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-334)

Add functionality to include crown court hearing results for trials. This includes 

- refactoring existing classes to remove mention of `magistrates`, as those classes can be reused for crown court results
- adding a new class to handle calculation of crown court outcome data

Not included here is handling crown court appeals. That functionality will be added in a separate PR.

## Checklist

### DIFF
--- a/app/services/crown_court_outcome_creator.rb
+++ b/app/services/crown_court_outcome_creator.rb
@@ -23,22 +23,22 @@ class CrownCourtOutcomeCreator < ApplicationService
   private
 
   def result_is_a_conclusion?
-    defendant&.dig(:offences)&.any? { |offence| offence.dig(:verdict).present? }
+    defendant.dig(:offences)&.any? { |offence| offence.dig(:verdict).present? }
   end
 
   def trial_outcome
-    return 'CONVICTED' if defendant&.dig(:offences)&.all? do |offence|
+    return 'CONVICTED' if defendant.dig(:offences)&.all? do |offence|
       GUILTY_VERDICTS.include? offence.dig(:verdict, :verdictType, :categoryType)
     end
-    return 'PART CONVICTED' if defendant&.dig(:offences)&.any? do |offence|
+    return 'PART CONVICTED' if defendant.dig(:offences)&.any? do |offence|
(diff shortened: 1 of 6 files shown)


app/services/crown_court_outcome_creator.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def trial_outcome
+    return 'CONVICTED' if defendant&.dig(:offences)&.all? do |offence|
+      GUILTY_VERDICTS.include? offence.dig(:verdict, :verdictType, :categoryType)
+    end
+    return 'PART CONVICTED' if defendant&.dig(:offences)&.any? do |offence|
+      GUILTY_VERDICTS.include? offence.dig(:verdict, :verdictType, :categoryType)
+    end
+
+    'AQUITTED'
```
archferns, line 37: Not sure if this was intentional for compatibility or just a typo.
```suggestion
    'ACQUITTED'
```

app/services/crown_court_outcome_creator.rb
```diff
@@ -0,0 +1,45 @@
...
+      ccooOutcome: trial_outcome,
+      caseEndDate: case_end_date,
+      appealType: nil
+    }
+  end
+
+  private
+
+  def result_is_a_conclusion?
+    defendant&.dig(:offences)&.any? { |offence| offence.dig(:verdict).present? }
```
archferns, line 26: Since defendant is a required param, I'm wondering if perhaps we could live without the safe navigation on `defendant.dig`?
```suggestion
    defendant.dig(:offences)&.any? { |offence| offence.dig(:verdict).present? }
```

app/services/crown_court_outcome_creator.rb
mpw5: Unfortunately that's how it's spelled in the MAAT database, so we need to keep it like that for compatibility...

image

I've added a comment explaining this though, so that it doesn't get accidentally 'fixed' in future.

app/services/crown_court_outcome_creator.rb
mpw5: Good point, fixed 👍

PR #184, opened 2020-06-09
Cacp 244 versioning oauth endpoints

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=398&projectKey=CACP&modal=detail&selectedIssue=CACP-244

Versioning of OAUTH Endpoints 
-Versioning of routes provided by doorkeeper gem 
-Creation of authorization, token info and tokens controllers 
-Updated RSPEC test suite accordingly

This ensures that future changes will not break current functionality. 

## Checklist

### DIFF
--- a/spec/routing/v1/doorkeeper/authorize_routing_spec.rb
+++ b/spec/routing/v1/doorkeeper/authorize_routing_spec.rb
@@ -1,0 +1,40 @@
+RSpec.describe 'routing', type: :routing do
+  it 'routes to #new' do
+    expect(get: '/v1/oauth/authorize').to route_to('v1/doorkeeper/authorizations#new')
+  end
+
+  it 'routes to #show' do
+    expect(get: '/v1/oauth/authorize/native').to route_to('v1/doorkeeper/authorizations#show')
+  end
+
+  it 'routes to #destroy' do
+    expect(delete: '/v1/oauth/authorize/').to route_to('v1/doorkeeper/authorizations#destroy')
+  end
+
+  it 'routes to #create' do
+    expect(post: '/v1/oauth/authorize').to route_to('v1/doorkeeper/authorizations#create')
+  end
+
+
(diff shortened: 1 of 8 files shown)


spec/routing/v1/doorkeeper/authorize_routing_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  it 'routes to #new' do
+    expect(get: '/v1/oauth/authorize').to route_to('v1/doorkeeper/authorizations#new')
+  end
+
+  it 'routes to #show' do
+    expect(get: '/v1/oauth/authorize/native').to route_to('v1/doorkeeper/authorizations#show')
+  end
+
+  it 'routes to #destroy' do
+    expect(delete: '/v1/oauth/authorize/').to route_to('v1/doorkeeper/authorizations#destroy')
```
mpw5, line 13: I don't think you need the trailing slash on `/v1/oauth/authorize/`. `/v1/oauth/authorize` would be more consistent with the other paths.

spec/routing/v1/doorkeeper/authorize_routing_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  it 'routes to #create' do
+    expect(post: '/v1/oauth/authorize').to route_to('v1/doorkeeper/authorizations#create')
+  end
+
+  context 'when no API version is specified' do
+    it 'routes to #create' do
+      expect(get: '/oauth/authorize').to route_to('v1/doorkeeper/authorizations#new')
+    end
```
mpw5, line 23: I think this is a duplicate of the test on lines 29-31. You can remove one of them. 

If you keep this one then the description should to be changed to `it 'routes to #new' do`, as you already have a test for `#create` on line 37.

PR #205, opened 2020-06-25
Amend pg default table access and default oids use

jsugarman:
## What
Amend pg default table access and default oids use

## Why
It maybe that I am hitting issues because I am using
postgres 9.6.15 locally, but when i run specs i get errors.

```
psql:../laa-court-data-adaptor/db/structure.sql:28: ERROR:  unrecognized configuration parameter "default_table_access_method"
rails aborted!
...
Tasks: TOP => db:test:load_structure
(See full trace by running task with --trace)
Migrations are pending. To resolve this issue, run:

        rails db:migrate RAILS_ENV=test
No examples found.
```
when i run `rails db:migrate RAILS_ENV=test` it amends
the db/structure.sql as in this commit.

## Checklist

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -10,6 +10,20 @@ SET client_min_messages = warning;
 SET row_security = off;
 
 --
+-- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
+
+
+--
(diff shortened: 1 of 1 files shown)


archferns:
It is definitely due to the postgres version, we are using v11.5.
The changes to structure.sql will get overwritten in a new migration though.

archferns:
Closing this as a [recent pg db version is recommended](https://github.com/ministryofjustice/laa-court-data-adaptor#set-up)

PR #212, opened 2020-07-02
Cacp 410 additional hearing info defence advocate

chris-groves:
## What

https://dsdmoj.atlassian.net/browse/CACP-410

Return Defence Advocate details - first name and last name, so they are ready to be displayed via the View Court Data application

## Checklist

### DIFF
--- a/app/models/hearing.rb
+++ b/app/models/hearing.rb
@@ -40,7 +40,7 @@ class Hearing < ApplicationRecord
   def defence_advocate_names
     defence_details = hearing_body['defenceCounsels']
     defence_details.map do |detail|
-	     "#{detail['firstName']} #{detail['lastName']}"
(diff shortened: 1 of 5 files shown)


app/models/hearing.rb
```diff
@@ -51,6 +57,10 @@ def defendants
     prosecution_cases.flat_map { |prosecution_case| prosecution_case['defendants'] }
   end
 
+  def defence_details
+    hearing_body['defenceCounsels']
```
archferns, line 61: As with prosecutionCaseCounsels, we might need to consider a safe navigation via `dig` or similar.
https://github.com/ministryofjustice/laa-court-data-adaptor/blob/56292ca03065155b3325cae7c2163f325a87b554/app/models/hearing.rb#L35-L37

PR #215, opened 2020-07-03
Cacp 410 additional hearing info defence advocate

chris-groves:
## What

https://dsdmoj.atlassian.net/browse/CACP-410

Implement ability to process Defence Advocate first name and last name, ready to be displayed by View Court Data application.

## Checklist

### DIFF
--- a/app/models/hearing.rb
+++ b/app/models/hearing.rb
@@ -40,7 +40,7 @@ class Hearing < ApplicationRecord
   def defence_advocate_names
     defence_details = hearing_body['defenceCounsels']
     defence_details.map do |detail|
-	     "#{detail['firstName']} #{detail['lastName']}"
+      "#{detail['firstName']} #{detail['lastName']}"
     end
   end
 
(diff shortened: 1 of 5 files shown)


app/models/hearing.rb
```diff
@@ -51,6 +57,10 @@ def defendants
     prosecution_cases.flat_map { |prosecution_case| prosecution_case['defendants'] }
   end
 
+  # def defence_details
+  #   hearing_body['defenceCounsels']
+  # end
+
```
archferns, line 63:
```suggestion
```

spec/models/hearing_spec.rb
```diff
@@ -51,6 +51,7 @@
 
       it { expect(hearing.judge_names).to eq(['Mr Recorder J Patterson']) }
       it { expect(hearing.prosecution_advocate_names).to eq(['John Rob']) }
+      it { expect(hearing.defence_advocate_names).to eq(['John Rob']) }
```
archferns, line 54: Just a nitpick: it would be ideal to have a different name than the `prosecution_advocate_names`

PR #235, opened 2020-07-22
Cacp 372 providers attending hearing info

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=398&modal=detail&selectedIssue=CACP-372

Create relationship between the Hearing model and the Provider model as a means to expose provider data

## Checklist


spec/fixtures/files/provider.json
```diff
@@ -0,0 +1,14 @@
+            {
```
archferns, line 1: Not a blocker: Lets indent this to the standard 2 spaces.

PR #245, opened 2020-07-31
Enable inclusions on defendant endpoint

jsugarman:
## What
Enable the retrieval of offences as part of
a defendant query

## Why
VCD is to query defendants on this endpoint and
display their offences on the same view/page. One
call to include all data is therefore more performant.

## Checklist

### DIFF
--- a/app/controllers/api/internal/v1/defendants_controller.rb
+++ b/app/controllers/api/internal/v1/defendants_controller.rb
@@ -48,7 +48,7 @@ module Api
         end
 
         def serialization_options
-          return { include: inclusions, params: { inclusions: inclusions } } if params[:include].present?
+          return { include: inclusions } if params[:include].present?
 
           {}
         end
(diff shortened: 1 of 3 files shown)


app/controllers/api/internal/v1/defendants_controller.rb
```diff
@@ -46,6 +46,16 @@ def allowed_params
         def transformed_params
           @transformed_params ||= unlink_params.slice(*allowed_params).to_hash.transform_keys(&:to_sym).merge(defendant_id: params[:id])
         end
+
+        def serialization_options
+          return { include: inclusions, params: { inclusions: inclusions } } if params[:include].present?
```
archferns, line 51: Not sure if we need this fine grained control over sending the params inside.
The reason this has been added to case search is  _if_ the UI requests a hearing to be included (additional calls need to be made then)
I assume this would have worked out of the box for your use case?

jsugarman: removed that now

PR #257, opened 2020-08-12
Dockerise app for development/testing

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-337)
Dockerise app to run both dev and specs using compose.

## Checklist

### DIFF
--- a/helm_deploy/laa-court-data-adaptor/templates/_environment.tpl
+++ b/helm_deploy/laa-court-data-adaptor/templates/_environment.tpl
@@ -14,6 +14,8 @@ env:
       secretKeyRef:
         name: ec-cluster-output
         key: url
+  - name: RAILS_ENV
+    value: 'production'
   - name: HOST_ENV
     value: {{ .Values.rails.host_env }}
   - name: SECRET_KEY_BASE
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,11 @@
 !/log/.keep
 !/tmp/.keep
 
+# Ignore pidfiles, but keep the directory.
+/tmp/pids/*
+!/tmp/pids/
+!/tmp/pids/.keep
+
 .byebug_history
 
 # Ignore master key for decrypting credentials and more.
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,8 +46,6 @@ RUN gem install bundler:2.1.4 -N \
 # DEPENDENCIES END #
 ####################
 
-ARG RAILS_ENV
-ENV RAILS_ENV ${RAILS_ENV:-production}
 ENV RAILS_SERVE_STATIC_FILES true
 ENV RAILS_LOG_TO_STDOUT true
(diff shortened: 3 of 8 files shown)


helm_deploy/laa-court-data-adaptor/templates/_environment.tpl
```diff
@@ -14,6 +14,8 @@ env:
       secretKeyRef:
         name: ec-cluster-output
         key: url
+  - name: RAILS_ENV
+    value: 'production'
```
richrace, line 18: Should this be hardcoded as `production`?

richrace:
Once this goes into master, we'll need to update any related Readmes for the new instructions to get this running locally. 

E.g. https://github.com/ministryofjustice/laa-court-data-ui/blob/master/docs/development.md

helm_deploy/laa-court-data-adaptor/templates/_environment.tpl
archferns: The idea was to run the cloud version in `production`, and control via related ENV vars, like `SENTRY_ENV` and so on. Happy to change to a different approach though.

archferns: Seeing a similar approach in [apply](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/9087b689771dcfe37db7ee4891be037a8ce5977a/helm_deploy/apply-for-legal-aid/templates/_envs.tpl#L72-L73) and [get paid CCCD](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/13a8140c35ba841d9172241fe0c99d1572d68b34/kubernetes_deploy/jobs/migrate.yaml#L20-L21)

richrace: The main reason why it looks odd to me, is that we have dev, test, uat, stage and production all in the Cloud Platform. So seeing this hardcoded makes me wonder what it is in those envs?

richrace: If this is the accepted LAA way of doing things, then that's okay.

archferns: > The main reason why it looks odd to me, is that we have dev, test, uat, stage and production all in the Cloud Platform. So seeing this hardcoded makes me wonder what it is in those envs?

About the dev, test, uat, stage and prod, the difference is defined via k8s secrets/helm values (eg:  SQS queues and API urls)

richrace: Yea, typically I'd see the real envs to match those here: https://github.com/ministryofjustice/laa-court-data-adaptor/tree/master/config/environments

This is definitely a different conversation.

PR #261, opened 2020-08-20
Cacp 475 remove offences without status date when ap

chris-groves:
## What

https://dsdmoj.atlassian.net/browse/CACP-475

Amend validations to make statusDate and effectiveStartDate optional when the statusCode is AP when CDA receives data from MAAT API

Filter offences in RepresentationOrderCreator service so only cases with a statusDate are sent to HMCTS Common Platform from CDA

## Checklist


archferns (approved):
:100: 
Nice work!

PR #274, opened 2020-08-28
Cacp 463 display if hearing resulted

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=398&modal=detail&selectedIssue=CACP-463

Show which hearings have resulted to avoid VCD making unnecessary requests for cases that have not resulted. 

## Checklist

### DIFF
--- a/app/models/hearing_summary.rb
+++ b/app/models/hearing_summary.rb
@@ -19,8 +19,8 @@ class HearingSummary
     body['hearingDays'].map { |hearing_day| hearing_day['sittingDay'] }
   end
 
-  def hearing_resulted?
-    Hearing.find_by(id: id) != nil
+  def resulted?
+    Hearing.find_by(id: id).present?
   end
 
   private
--- a/app/models/hearing_summary.rb
+++ b/app/models/hearing_summary.rb
@@ -19,6 +19,10 @@ class HearingSummary
     body['hearingDays'].map { |hearing_day| hearing_day['sittingDay'] }
   end
 
+  def hearing_resulted?
+    Hearing.find_by(id: id) != nil
+  end
+
   private
 
   def court_centre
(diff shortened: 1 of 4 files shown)


app/models/hearing_summary.rb
```diff
@@ -19,6 +19,10 @@ def hearing_days
     body['hearingDays'].map { |hearing_day| hearing_day['sittingDay'] }
   end
 
+  def hearing_resulted?
+    Hearing.find_by(id: id) != nil
```
archferns, line 23:
```suggestion
    Hearing.find_by(id: id).present?
```
This would be better served by a `present?` check perhaps.

spec/models/hearing_spec.rb
```diff
@@ -4,6 +4,10 @@
   let(:hearing) { described_class.new(body: '{}') }
   subject { hearing }
 
+  context 'the hearing has not been resulted' do
+    it { expect(subject).to_not be_persisted }
+  end
+
```
archferns, line 10: Not entirely sure if this adds value to the `resulted?` check itself, so I would suggest we take this off.

spec/models/hearing_spec.rb
```diff
@@ -27,6 +31,7 @@
     it { expect(hearing.providers).to be_empty }
     it { expect(hearing.provider_ids).to be_empty }
     it { expect(hearing.judge_names).to be_empty }
+    it { expect(hearing).to be_persisted }
```
archferns, line 34: As above

app/models/hearing_summary.rb
```diff
@@ -19,6 +19,10 @@ def hearing_days
     body['hearingDays'].map { |hearing_day| hearing_day['sittingDay'] }
   end
 
+  def hearing_resulted?
```
archferns, line 22: Minor nitpick: should we name this just `resulted?` to be able to use it as `hearing.resulted?`, rather than `hearing.hearing_resulted?`

archferns (approved):
LGTM

PR #298, opened 2020-09-28
Return a 404 for api/external routes when accessed by internal routes

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/CACP-342
Since external routes (CP) should be accessed only over mTLS, this change will return a 404 for non-mtls connections.

## Checklist


chris-groves (dismissed):

chris-groves (dismissed):

alexdesi:
This branch is outdated (almost 3yo), so I am closing it.

PR #310, opened 2020-10-12
Lasb 379 linking when no hearing

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=454&modal=detail&selectedIssue=LASB-379

Creating a link request when the hearing summary is not available results in a failure. This prevents a failure further down the line as linking a case in MLRA relies on the court location being present from the hearing summary.

## Checklist

### DIFF
--- a/spec/services/link_validator_spec.rb
+++ b/spec/services/link_validator_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe LinkValidator do
     end
   end
 
-  context 'when the prosecution case defendant offence does not exit' do
+  context 'when the prosecution case defendant offence does not exist' do
     before do
       ProsecutionCaseDefendantOffence.destroy_all
     end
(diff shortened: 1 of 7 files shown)


spec/services/link_validator_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      prosecution_case.body.delete('hearingSummary')
+      prosecution_case.save!
+    end
+
+    it 'returns false' do
+      expect(subject).to eq false
+    end
+  end
+
+  context 'when the prosecution case defendant offence does not exit' do
```
archferns, line 35:
```suggestion
  context 'when the prosecution case defendant offence does not exist' do
```

archferns (approved):
LGTM

archferns (approved):

PR #313, opened 2020-10-19
Pluralise provider serialization type

jsugarman:
## What
Pluralize the ProviderSerializer `type`

Currently type is not explictly set, resulting in the type being
automatically set to :provider
(singular). This breaks the pattern for other serializers that are plural.

The result was even when `hearings.providers` was "included" in a query,
and providers returned, the consumer (VCD) was unable to retrieve the
providers following that pattern.

## Checklist

### DIFF
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -2,5 +2,7 @@
 
 class ProviderSerializer
   include FastJsonapi::ObjectSerializer
+  set_type :providers
+
   attributes :name, :role
 end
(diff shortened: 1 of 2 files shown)


.env.test
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgres://localhost/laa_court_data_adaptor_test
```
archferns, line 1: we probably don't want this added here as this might interfere with the test setup

archferns (approved):
:+1:

PR #324, opened 2020-11-06
Expose mode of trial reason and reason code

jsugarman:
Expose mode of trial reason and reason code

## What
Expose mode of trial reason and reason code

Needed by VCD to display to case workers as it
it impacts claimable values.

[CBO-1575 display mode of trial reason](https://dsdmoj.atlassian.net/browse/CBO-1575)

## Checklist


archferns (approved):
Thank you :+1:

PR #340, opened 2020-12-01
Return unique pleas per offence

jsugarman:
## What
Prevent non-unique pleas being returned

[Link to story LASB-442](https://dsdmoj.atlassian.net/browse/LASB-442)

## Why

Pleas are repeated per hearing, defendant, offence on the hearing endpoint
of common platoform API. We therefore need to filter out non-unique plea
details.

This will enable consumers (VCD) to easily present Changes of plea without
pointless repetition.

## Checklist

### DIFF
--- a/app/models/offence.rb
+++ b/app/models/offence.rb
@@ -69,7 +69,7 @@ private
   def pleas_array
     return [] if details.blank?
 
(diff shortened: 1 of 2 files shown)


archferns (approved):
LGTM

PR #342, opened 2020-12-04
Modify offence model to expose multiple mode of trial reasons

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=454&projectKey=LASB&modal=detail&selectedIssue=LASB-443

Expose all the allocation decisions (mode of trial reasons and mode of trial reason codes) available against a single offence. Although it is expected that only one decision will be available, this accounts for the possibility that multiple decisions could be.

## Checklist

### DIFF
--- a/app/models/offence.rb
+++ b/app/models/offence.rb
@@ -37,6 +37,15 @@ class Offence
     allocation_decisions.dig(0, "motReasonCode")
   end
 
+  def mode_of_trial_reasons
+    allocation_decisions.map do |decision|
+      {
+        description: decision["motReasonDescription"],
+        code: decision["motReasonDescriptionCode"],
+      }
+    end
+  end
+
   def maat_reference
     laa_reference["applicationReference"] if laa_reference.present?
   end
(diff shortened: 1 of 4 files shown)


jsugarman (approved):
looks great to me 👍

archferns (approved):

app/models/offence.rb
```diff
@@ -75,6 +84,6 @@ def pleas_array
   def allocation_decisions
     return [] if details.blank?
 
-    details.flat_map { |detail| detail["allocationDecision"] }
+    details.flat_map { |detail| detail["allocationDecision"] }.uniq.compact
```
jsugarman, line 87: I believe this `uniq.compact` needs to go on the mode_of_trial_reasons block at line:46. 

This is because it may be possible for an  allocation decision to be unique du to ids, as below, but its code and description to not be

```
                "allocationDecision": {
                  "originatingHearingId": "75865225-3608-4715-a800-c8b001a7f6d2",
                  "offenceId": "a588aefd-006a-41d9-a42d-3beb450cbc0f",
                  "allocationDecisionDate": "2020-08-26",
                  "motReasonId": "f8eb278a-8bce-373e-b365-b45e939da38a",
                  "sequenceNumber": 30,
                  "motReasonDescription": "Defendant elects trial by jury",
                  "motReasonCode": "04"
                },
```

PR #349, opened 2020-12-11
Add ADR template

archferns:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-357)
Add ADRs to CDA

## Checklist

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+docs/architecture/decisions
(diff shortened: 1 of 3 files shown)


archferns:
I believe it is a matter of convention, to start the readme itself as 0001, and is supported out of the box by `adr-tools` which auto generates it by default. I'm happy to change, but my personal preference is the convention. https://github.com/npryce/adr-tools/blob/master/doc/adr/0001-record-architecture-decisions.md

arthurashman:
OK, good to know. Thank you for the info. Very happy to go with the convention.

arthurashman (approved):
Looks good to me

PR #354, opened 2020-12-18
Expose cracked ineffective trial payload

jsugarman:
## What
Expose cracked inneffective trial payload

Add `cracked_ineffective_trial` payloads to the
hearing and prosecution case endpoints.

## Ticket

[CBO-1605](https://dsdmoj.atlassian.net/browse/CBO-1605)

## Why
These are required by VCD to inform case workers if a trial
has cracked (or been vacated in a manner that would
constitute a cracked trial - TBC). This in turn affects claimable
fees and the amounts.

## TODO

 - [ ] check with crimeapps teams if this is the best way
 - [ ] check with crimeapps teams we need to generate any JSON for the new pseudo-model
 - [ ] check with crimeapps teams if we need more specs or swagger doc changes
 - [x]  rebase/tidy commits

## Checklist

### DIFF
--- a/app/models/cracked_ineffective_trial.rb
+++ b/app/models/cracked_ineffective_trial.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class CrackedIneffectiveTrial
+  include ActiveModel::Model
+
+  attr_accessor :body
+
+  def id
+    body["id"]
+  end
+
+  def code
+    body["code"]
+  end
+
+  def description
+    body["description"]
+  end
+
+  def type
+    body["type"]&.downcase
+  end
+end
--- a/app/models/hearing.rb
+++ b/app/models/hearing.rb
@@ -55,6 +55,16 @@ class Hearing < ApplicationRecord
     hearing_body["defenceCounsels"]&.map { |defence_counsel| defence_counsel["id"] }
   end
 
+  def cracked_ineffective_trial
+    return if cp_cracked_ineffective_trial.blank?
+
+    CrackedIneffectiveTrial.new(body: cp_cracked_ineffective_trial)
+  end
+
+  def cracked_ineffective_trial_id
+    cp_cracked_ineffective_trial&.fetch("id", nil)
+  end
+
 private
 
   def hearing_body
@@ -81,5 +91,9 @@ private
 
   def court_centre
     @court_centre ||= HmctsCommonPlatform::Reference::CourtCentre.find(hearing_body["courtCentre"]["id"])
+  end
+
+  def cp_cracked_ineffective_trial
+    hearing_body["crackedIneffectiveTrial"]
   end
 end
--- a/spec/models/hearing_spec.rb
+++ b/spec/models/hearing_spec.rb
@@ -82,5 +82,27 @@ RSpec.describe Hearing, type: :model do
         it { expect(hearing.defence_advocate_names).to be_nil }
       end
     end
+
+    context "with cracked ineffective trial data available" do
+      let(:hearing_id) { "da124701-048f-408c-85b4-81138316ddce" }
+      let(:hearing) do
+        VCR.use_cassette("hearing_result_fetcher/success_hearing_cracked_trial") do
+          Api::GetHearingResults.call(hearing_id: hearing_id)
+        end
+      end
+
+      describe "#cracked_ineffective_trial_id" do
+        subject { hearing.cracked_ineffective_trial_id }
+
+        it { is_expected.to eql("c4ca4238-a0b9-3382-8dcc-509a6f75849b") }
+      end
+
+      describe "#cracked_ineffective_trial" do
+        subject { hearing.cracked_ineffective_trial }
+
+        it { is_expected.to be_a(CrackedIneffectiveTrial) }
+        it { is_expected.to respond_to(:id, :code, :description, :type) }
+      end
+    end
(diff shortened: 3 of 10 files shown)


jsugarman:
@archferns hi, have PR'd this is the knowledge that you are in next week and it is your laster week 😬 . I am on leave that week so can you review this and provide any suggestions as to how, if different, it should be done, or things i may have missed.

please see the TODO list for more of a breakdown of what i think would be good if you ad @chris-groves could check/confirm

app/models/hearing.rb
```diff
@@ -55,6 +55,16 @@ def provider_ids
     hearing_body["defenceCounsels"]&.map { |defence_counsel| defence_counsel["id"] }
   end
 
+  def cracked_ineffective_trial
+    return if cp_cracked_ineffective_trial.blank?
+
```
chris-groves, line 60: could 'cp' be written out in full for extra clarity?

chris-groves (dismissed):

app/models/hearing.rb
```diff
@@ -55,6 +55,16 @@ def provider_ids
...
+    CrackedIneffectiveTrial.new(body: cp_cracked_ineffective_trial)
+  end
+
+  def cracked_ineffective_trial_id
+    cp_cracked_ineffective_trial&.fetch("id", nil)
+  end
```
chris-groves, line 66: see earlier comment about 'cp'

app/models/hearing.rb
```diff
@@ -82,4 +92,8 @@ def hearing_day_events
   def court_centre
     @court_centre ||= HmctsCommonPlatform::Reference::CourtCentre.find(hearing_body["courtCentre"]["id"])
   end
+
+  def cp_cracked_ineffective_trial
+    hearing_body["crackedIneffectiveTrial"]
```
chris-groves, line 97: see earlier comment about 'cp'

app/models/hearing.rb
jsugarman: Thanks @chris-groves . Normally i'd agree but `cp` abbreviation for common platform is used in a fair few  places (list below) in the codebase and is a relatively well known in the context of this repo. I personally prefer the abbreviation but what does @arthurashman @msahilan07 think?

If we change it we should probably be consistent and remove it from other places it is used to.

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/maat_link_creator.rb#L10

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/maat_link_creator.rb#L23

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/representation_order_creator.rb#L13

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/representation_order_creator.rb#L18

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/laa_reference_unlinker.rb#L15

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/laa_reference_unlinker.rb#L23

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/defendant_finder.rb#L10

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/services/defendant_finder.rb#L15

...

app/models/cracked_ineffective_trial.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class CrackedIneffectiveTrial
+  include ActiveModel::Model
+
+  attr_accessor :body
+
+  def id
+    body["id"]
```
arthurashman, line 9: Would there be any sensible clarifying nouns to goes with these values, similar to in the offence model? https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/models/offence.rb

app/models/cracked_ineffective_trial.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def code
+    body["code"]
```
arthurashman, line 13: Likewise here

app/models/hearing.rb
arthurashman: My preference would probably be to write this out, but maybe we should do all instances as part of a different ticket in the future separate from this PR

spec/models/hearing_spec.rb
```diff
@@ -5,69 +5,76 @@
 
   let(:hearing) { described_class.new(body: "{}") }
 
-  describe "validations" do
-    it { is_expected.to validate_presence_of(:body) }
-  end
+  it { is_expected.to validate_presence_of(:body) }
 
-  describe "Common Platform search" do
+  context "when searching common platform" do
```
arthurashman, line 10: This is much better, thank you

spec/models/hearing_spec.rb
```diff
@@ -5,69 +5,76 @@
 
   let(:hearing) { described_class.new(body: "{}") }
 
-  describe "validations" do
-    it { is_expected.to validate_presence_of(:body) }
-  end
+  it { is_expected.to validate_presence_of(:body) }
```
arthurashman, line 8: Would this be your recommended way to do this outside of any describe block? It is like this in a few places. We can put it on a backlog to change at some point if this is preferred.

arthurashman (approved):

PR #355, opened 2020-12-18
Ensure only unique mode of trial reasons are returned

chris-groves:
## What
Ensure only unique mode of trial reasons are returned.

[LASB-505](https://dsdmoj.atlassian.net/browse/LASB-505)
[extension of LASB-443](https://dsdmoj.atlassian.net/browse/LASB-443)

## Why
Ensure only unique mode of trial reasons are returned.

Because `allocationDecisions` appear to be repeating
across hearings for the same offence (unlike on
the  SIT environment). The `originatingHearingId` for the
allocationDecision hash will be unique even if it
is for the same offenceId and the same mode of trial reason;
a chain of `uniq` on the entire allocationDecision will 
therefore not do what is intended.

## Checklist

### DIFF
--- a/app/models/offence.rb
+++ b/app/models/offence.rb
@@ -66,6 +66,6 @@ private
   def allocation_decisions
     return [] if details.blank?
 
-    details.flat_map { |detail| detail["allocationDecision"] }.uniq.compact
+    details.flat_map { |detail| detail["allocationDecision"] }
   end
 end
--- a/app/models/offence.rb
+++ b/app/models/offence.rb
@@ -30,12 +30,12 @@ class Offence
   end
 
   def mode_of_trial_reasons
-    allocation_decisions.map do |decision|
+    allocation_decisions.map { |decision|
       {
         description: decision["motReasonDescription"],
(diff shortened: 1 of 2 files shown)


app/models/offence.rb
```diff
@@ -38,12 +38,12 @@ def mode_of_trial_reason_code
...
 
   def mode_of_trial_reasons
-    allocation_decisions.map do |decision|
+    allocation_decisions.map { |decision|
       {
         description: decision["motReasonDescription"],
         code: decision["motReasonCode"],
       }
-    end
+    }.uniq.compact
```
jsugarman, line 46: 👍

jsugarman (changes requested):
I think we should remove the uniq.compact from line 77 of`app/models/offence.rb` as well.

Also, we should test this repeated allocation decision scenario if it is not already  - I can't see a test for this.

jsugarman (approved):
Have amended but would be good to have someone else review.

This should go out asap as it is displaying odd details ATM

msahilan07 (approved):
LGTM

PR #372, opened 2021-01-11
Silence spec timings

jsugarman:
## What
Silence spec timing output

## Why
A preference but they are rather noisy and can
be switched on if and when we need to think
about improving test suite speed, which is not a
problem at the moment.

## Checklist


arthurashman (approved):
Agreed

PR #374, opened 2021-01-11
Refactor contexts and examples hearing spec

jsugarman:
## What
Refactor contexts and examples hearing spec

I have amended naming to include the use of when/with/but
in the contexts (as recommedned by https://www.betterspecs.org/)
and tested the same attributes in the different contexts for
consistency.

## Why
Found these specs unclear and a little inconsistent
when working on a related ticket.

## Checklist


jsugarman:
needs to be part of the cracked ineffective trial PR

PR #381, opened 2021-01-15
change json serializer gem, change include statements in serializers

arthurashman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-LASB448)

Replaced fast_jsonapi gem as it is no longer maintained. 
Replaced with jsonapi-serializer gem which is an open-source project forked from fast_jsonapi. 
Preference using this over Rails default serializer on the assumption that there is benefit to the performance gains.
Serializers updated to use the correct include statements

## Checklist


arthurashman:
Speed testing shows comparable results

chris-groves (approved):

PR #383, opened 2021-01-21
Add ping endpoint

jsugarman:
## What
Add ping endpoint to Adaptor

#### Ticket 
[CBO-1668](https://dsdmoj.atlassian.net/browse/CBO-1668)

## Why

So we can easily retrieve versioning/release related
information for each environment on which the adaptor
is deployed.

## Example usage of ping

    e.g. A shell script to query version info for a list of ingresses
         requires [jq](https://stedolan.github.io/jq/))

```
function pingit {
  for i in $@;
  do
    uri="https://${i}/ping.json";
    echo $uri;
    response=`curl --silent $uri`;
    echo $response | jq .;
    echo;
  done
}
```

And use:
```
$ pingit <list-of-ingresses>
````

## Checklist


kmahern (approved):
reviewed on test environment and looks good.

Issue #395, opened 2021-02-04
defendant name is adding additional whitespace


jsugarman:
**Expected behavior** 
When we call a defendants name attribute it is expected to be single space delimited
e.g. `Jammy Dodger`

**Actual behaviour**
The name is double spaced delimited
e.g. `Jammy  Dodger`

**Cause**

https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/models/defendant.rb#L13
```
[first_name, middle_name, last_name].compact.join(" ")
```
when middle name is an empty string it results in an extra whitespace.
```
['Jammy', '', 'Dodger'].compact.join(" ")
=> "Jammy  Dodger"
```

**Impact** 
Just noticed that VCD searching by name does not work as a result unless you put two spaces between first and last name you are search for

**Fix**
use  rails's `squish` or ruby's `squeeze(' ')`
```
['Jammy', '', 'Dodger'].join(' ').squish
 => "Jammy Dodger"

# ALSO WORKS WITH NILS
['Jammy', nil, 'Dodger'].join(' ').squish
 => "Jammy Dodger"
```

laurentqro:
Fixed in https://github.com/ministryofjustice/laa-court-data-adaptor/pull/397.

PR #397, opened 2021-02-05
Remove extraneous whitespace from defendant.name attribute

jsugarman:
## What
Remove extraneous whitespace from defendant.name attribute

## Why
Extra whitespace is added for empty middle names
resulting in search failure from VCD as the
name search is currently an extact case-insensitive
match. Whitespace is not currently handled.

This relates to the [issue 395](https://github.com/ministryofjustice/laa-court-data-adaptor/issues/395) that I have raised:

VCD is also providing a [belt and braces fix](https://github.com/ministryofjustice/laa-court-data-ui/pull/456):

## Checklist

### DIFF
--- a/app/models/defendant.rb
+++ b/app/models/defendant.rb
@@ -10,7 +10,7 @@ class Defendant
   end
 
   def name
-    [first_name, middle_name, last_name].compact.join(" ")
+    [first_name, middle_name, last_name].join(" ").squish
   end
 
   def first_name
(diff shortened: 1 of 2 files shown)


arthurashman (approved):
Thanks for this. Looks good. 

I know .squish will remove the extra whitespaces, but would it be worth having a test where the middle name is a whitespace?

laurentqro (approved):

jsugarman:
> I know .squish will remove the extra whitespaces, but would it be worth having a test where the middle name is a whitespace?

the scenario that is most common by the looks of it is when middlename is an empty string, `""` and `join` adds the whitespace as a result - which i added a separate test for. I will amend the third test (fred bloggs) to include a whitespace in the middlename, as well as the first and last name but don't think it needs a seaprate test?!

arthurashman (approved):

PR #410, opened 2021-02-16
Include prosecution concluded flag in payload

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-531)

This PR makes sure CDA parses the prosecution concluded flag that is present on the defendant objects under a hearing and includes it as part of the payload that is posted to the queue for MAAT API to consume and process.

## Why

Posting the prosecution concluded flag allows MAAT API to process the case conclusion and determine the crown court outcome.

## Checklist

### DIFF
--- a/app/services/sqs/publish_hearing.rb
+++ b/app/services/sqs/publish_hearing.rb
@@ -52,6 +52,7 @@ module Sqs
         caseCreationDate: shared_time.to_date.strftime("%Y-%m-%d"),
         cjsLocation: cjs_location,
         docLanguage: "EN",
+        proceedingsConcluded: defendant.dig(:proceedingsConcluded),
         inActive: inactive?,
         defendant: defendant_hash,
         session: session_hash,
--- a/spec/services/sqs/publish_hearing_spec.rb
+++ b/spec/services/sqs/publish_hearing_spec.rb
@@ -26,6 +26,7 @@ RSpec.describe Sqs::PublishHearing do
   let(:defendant) do
     {
       'id': "c6cf04b5-901d-4a89-a9ab-767eb57306e4",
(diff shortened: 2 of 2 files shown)


msahilan07 (commented):
This looks good to me.

msahilan07 (approved):

app/services/sqs/publish_hearing.rb
```diff
@@ -52,6 +52,7 @@ def message
         caseCreationDate: shared_time.to_date.strftime("%Y-%m-%d"),
         cjsLocation: cjs_location,
         docLanguage: "EN",
+        proceedingsConcluded: defendant.dig(:proceedingsConcluded),
```
arthurashman, line 55: As we're not going into nested hashes, would ` &.` be preferable over `.dig` ?

arthurashman (approved):
As much a learning point for me as a requested change

chris-groves (approved):

app/services/sqs/publish_hearing.rb
laurentqro: `&.` is an operator that works like `try`, https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby, so it's not particular to hashes. `dig` was designed to fetch values in nested hashes, but can also be used as a synonym to `[]`. I do agree with you that `[]` might be better, in the sense that it's more intention revealing.

I'd suggest staying away from `&.` as much as possible, because it often violates the principle of least knowledge that is best practice in object-oriented design (aka "law of Demeter"), and the "fail fast" philosophy (https://martinfowler.com/ieeeSoftware/failFast.pdf).

PR #417, opened 2021-02-24
Fix undefined method `deep_transform_keys' for String

laurentqro:
We are calling deep_transform_keys (a Hash method), on the hearing body
from the database. However, hearing body is stored in Postgres as a JSONB
string, not a hash.

Tests were not picking up on this because our tests were assuming hearing body was
a hash.

https://sentry.io/organizations/ministryofjustice/issues/2093600232/?cursor=0%3A100%3A0&environment=prod&project=5375870

## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-550)

### DIFF
--- a/app/services/hearings_creator.rb
+++ b/app/services/hearings_creator.rb
@@ -2,7 +2,7 @@
 
 class HearingsCreator < ApplicationService
   def initialize(hearing_id:)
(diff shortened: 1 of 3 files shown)


laurentqro:
done here https://github.com/ministryofjustice/laa-court-data-adaptor/pull/419

PR #419, opened 2021-02-24
temporary patch to process failing delayed jobs

arthurashman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-LASB550)

Patch to fix undefined method error `deep_transform_keys' for String

Temporary method added to process failing delayed jobs

To be reverted once backlog or retried jobs has run

## Checklist


laurentqro (approved):
🚀

PR #424, opened 2021-02-26
Prevent invalid hearing body from being persisted or published to the queue

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-552)

https://sentry.io/organizations/ministryofjustice/issues/2237049764/events/8b17d4493642498d9ca4646687722086/?environment=prod&project=5375870

## Checklist

### DIFF
--- a/app/controllers/api/internal/v1/laa_references_controller.rb
+++ b/app/controllers/api/internal/v1/laa_references_controller.rb
@@ -5,9 +5,8 @@ module Api
     module V1
       class LaaReferencesController < ApplicationController
         def create
-          contract = NewLaaReferenceContract.new.call(**transformed_params)
           if contract.success?
-            LaaReferenceCreator.call(defendant_id: transformed_params[:defendant_id], user_name: transformed_params[:user_name], maat_reference: transformed_params[:maat_reference])
+            create_laa_reference
             render status: :accepted
           else
             render json: contract.errors.to_hash, status: :bad_request
@@ -15,6 +14,16 @@ module Api
         end
 
       private
+
+        def contract
+          @contract ||= NewLaaReferenceContract.new.call(**transformed_params)
+        end
+
+        def create_laa_reference
+          LaaReferenceCreator.call(defendant_id: transformed_params[:defendant_id],
+                                   user_name: transformed_params[:user_name],
+                                   maat_reference: transformed_params[:maat_reference])
+        end
 
         def create_params
           params.from_jsonapi.require(:laa_reference).permit(allowed_params)
--- a/app/services/hearing_recorder.rb
+++ b/app/services/hearing_recorder.rb
@@ -11,6 +11,14 @@ class HearingRecorder < ApplicationService
     if hearing_contract.success?
       hearing.update!(body: body)
       publish_hearing_to_queue if publish_to_queue
+    else
+      Sentry.capture_message("A hearing contract failed",
+                             tags: { hearing_id: hearing.id },
+                             extras: {
+                               body: body,
+                               publish_to_queue: publish_to_queue,
+                               errors: hearing_contract.errors.messages,
+                             })
     end
 
     hearing
--- a/app/services/hearing_recorder.rb
+++ b/app/services/hearing_recorder.rb
@@ -11,16 +11,6 @@ class HearingRecorder < ApplicationService
     if hearing_contract.success?
       hearing.update!(body: body)
       publish_hearing_to_queue if publish_to_queue
-    else
-      Sentry.capture_message("A hearing contract failed",
-                            tags: {
-                              hearing_id: hearing.id,
-                            },
-                            extras: {
-                              body: body,
-                              publish_to_queue: publish_to_queue,
-                              errors: hearing_contract.errors.messages,
-                            })
     end
 
     hearing
--- a/app/services/hearing_recorder.rb
+++ b/app/services/hearing_recorder.rb
@@ -8,13 +8,33 @@ class HearingRecorder < ApplicationService
   end
 
   def call
-    hearing.update!(body: body)
-    HearingsCreatorWorker.perform_async(Current.request_id, hearing.id) if publish_to_queue
+    if hearing_contract.success?
+      hearing.update!(body: body)
+      publish_hearing_to_queue if publish_to_queue
+    else
+      Sentry.capture_message("A hearing contract failed",
+                            tags: {
+                              hearing_id: hearing.id,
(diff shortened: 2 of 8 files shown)


app/services/hearing_recorder.rb
```diff
@@ -8,13 +8,33 @@ def initialize(hearing_id:, body:, publish_to_queue:)
   end
 
   def call
-    hearing.update!(body: body)
-    HearingsCreatorWorker.perform_async(Current.request_id, hearing.id) if publish_to_queue
+    if hearing_contract.success?
+      hearing.update!(body: body)
+      publish_hearing_to_queue
+    else
+      Raven.capture_message("A hearing contract failed",
```
arthurashman, line 15: Are you happy this will need changing for the sentry-ruby gem? Or is this worth changing now and merging the gem change first?

app/controllers/api/internal/v1/laa_references_controller.rb
```diff
@@ -16,6 +15,16 @@ def create
...
+        end
+
+        def create_laa_reference
+          LaaReferenceCreator.call(defendant_id: transformed_params[:defendant_id],
+                                   user_name: transformed_params[:user_name],
+                                   maat_reference: transformed_params[:maat_reference])
+        end
+
         def create_params
           params.from_jsonapi.require(:laa_reference).permit(allowed_params)
```
arthurashman, line 29: Would it be worth standardising the require/permit statements? They alter in format between this and the hearings controller

app/services/hearing_recorder.rb
```diff
@@ -8,13 +8,33 @@ def initialize(hearing_id:, body:, publish_to_queue:)
   end
 
   def call
-    hearing.update!(body: body)
-    HearingsCreatorWorker.perform_async(Current.request_id, hearing.id) if publish_to_queue
+    if hearing_contract.success?
+      hearing.update!(body: body)
+      publish_hearing_to_queue
```
arthurashman, line 13: This is less clear to me now. It looks like it will publish to queue even when `!publish_to_queue`

app/services/laa_reference_creator.rb
```diff
@@ -10,15 +10,22 @@ def initialize(defendant_id:, user_name: nil, maat_reference: nil)
   end
 
   def call
-    create_laa_reference!
-    MaatLinkCreatorWorker.perform_async(Current.request_id, laa_reference.id)
+    persist_laa_reference!
```
arthurashman, line 13: I'm interested in the argument for changing the name of this method.

app/services/maat_link_creator.rb
```diff
@@ -6,31 +6,41 @@ def initialize(laa_reference_id:)
   end
 
   def call
-    push_to_sqs unless laa_reference.dummy_maat_reference?
-    call_common_platform_endpoint
-    PastHearingsFetcherWorker.perform_at(30.seconds.from_now, Current.request_id, prosecution_case_id)
+    publish_laa_reference_to_queue
```
arthurashman, line 9: This is less clear to me now. It looks like it will publish to queue even when `dummy_maat_reference`

app/services/maat_link_creator.rb
```diff
@@ -6,31 +6,41 @@ def initialize(laa_reference_id:)
...
+    post_laa_references_to_common_platform
```
arthurashman, line 10: Lovely renaming

app/services/maat_link_creator.rb
```diff
@@ -6,31 +6,41 @@ def initialize(laa_reference_id:)
...
+    fetch_past_hearings
```
arthurashman, line 11: Lovely extracting to method and clear rename

app/services/maat_link_creator.rb
```diff
@@ -6,31 +6,41 @@ def initialize(laa_reference_id:)
...
   end
 
 private
 
-  def push_to_sqs
+  def publish_laa_reference_to_queue
+    return if laa_reference.dummy_maat_reference?
+
     Sqs::PublishLaaReference.call(defendant_id: laa_reference.defendant_id,
```
arthurashman, line 19: Any thoughts on renaming the Sqs module, so we're not thinking in terms of AWS in case we use a different queue service in the future? Different question for another time, but just a thought

spec/services/hearing_recorder_spec.rb
```diff
@@ -54,4 +54,51 @@
...
+
+      expect(hearing.reload.body).to eq({ "foo" => "bar" })
+    end
+
+    it "does not publish hearing to the queue" do
+      expect(HearingsCreatorWorker).not_to receive(:perform_async)
+    end
+  end
+
+  context "when the hearing body is jibberish" do
```
arthurashman, line 85: 😂

spec/services/hearing_recorder_spec.rb
```diff
@@ -54,4 +54,51 @@
...
+    let(:hearing) { Hearing.create!(body: { foo: "bar" }) }
+    let(:body) { "<span>Clearly not a processable hearing body</span>" }
```
arthurashman, line 87: Amen

arthurashman (dismissed):
Looks really solid. Happy that I've seen the fix and understand it. I've left a couple of comments that I'd be interested in hearing your thoughts on though.

app/services/maat_link_creator.rb
laurentqro: Totally agree, our code should not change if we decided to switch queueing providers 👌🏻

chris-groves (dismissed):

app/services/laa_reference_creator.rb
laurentqro: We're inside a class called `LaaReferenceCreator`, which does a bunch of things, so I wanted to be a bit more specific here.

app/services/maat_link_creator.rb
laurentqro: I put it back in 👌🏻

arthurashman: I don't understand what has been put back in. I was thinking it wants to be 
`publish_laa_reference_to_queue unless laa_reference.dummy_maat_reference?`

arthurashman (dismissed):

app/services/maat_link_creator.rb
laurentqro: sorry, mixed up two things

arthurashman (approved):

PR #425, opened 2021-02-26
Migrate to sentry-ruby from sentry-raven

laurentqro:
## What

[sentry-raven](https://github.com/getsentry/sentry-ruby/tree/master/sentry-raven) has entered maintenance mode, which means it won't receive any new feature supports or aggressive bug fixes.

This PR replaces it with [sentry-ruby](https://github.com/getsentry/sentry-ruby/tree/master/sentry-ruby).

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -305,8 +305,13 @@ GEM
     rubocop-rspec (1.42.0)
       rubocop (>= 0.87.0)
     ruby-progressbar (1.10.1)
-    sentry-raven (3.1.1)
+    sentry-ruby (4.2.2)
+      concurrent-ruby (~> 1.0, >= 1.0.2)
       faraday (>= 1.0)
+      sentry-ruby-core (= 4.2.2)
+    sentry-ruby-core (4.2.2)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -305,8 +305,13 @@ GEM
     rubocop-rspec (1.42.0)
       rubocop (>= 0.87.0)
     ruby-progressbar (1.10.1)
-    sentry-raven (3.1.1)
+    sentry-ruby (4.2.2)
```
arthurashman, line 308: We don't need any of the following?
gem "sentry-rails"
gem "sentry-sidekiq"
gem "sentry-delayed_job"

arthurashman (approved):

chris-groves (approved):

Gemfile.lock
laurentqro: My understanding is that these gems (that depend on sentry-ruby) bring a few little extras, but I don't think we need them. Could be interesting to look into how they could be useful to us.

PR #427, opened 2021-02-26
Fix undefined method positive? for nil:NilClass

laurentqro:
## What

Fixes issue https://github.com/ministryofjustice/laa-court-data-adaptor/issues/426.

## Checklist

### DIFF
--- a/app/services/api/search_prosecution_case.rb
+++ b/app/services/api/search_prosecution_case.rb
@@ -7,22 +7,28 @@ module Api
     end
 
     def call
-      record_search_results if successful_response?
+      record_prosecution_cases if search_results?
     end
 
   private
 
-    def successful_response?
-      response.status == 200 && response.body["totalResults"].positive?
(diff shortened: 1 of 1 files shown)


arthurashman (approved):
The fixing change is just adding the safe navigation operator?

Would it be possible in future to separate out the fix and the refactor to 2 PRs to make the fixing change more clear?

app/services/api/search_prosecution_case.rb
```diff
@@ -7,22 +7,28 @@ def initialize(params)
...
     def call
-      record_search_results if successful_response?
+      record_prosecution_cases if search_results?
     end
 
   private
 
-    def successful_response?
-      response.status == 200 && response.body["totalResults"].positive?
+    def search_results?
```
arthurashman, line 15: `search_results_successful?` ?

chris-groves (approved):

PR #430, opened 2021-03-01
Add sentry context

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-548)

## Checklist


arthurashman (approved):
Is it only when a contract fails that we tend to see Sentry logs?

laurentqro:
> Is it only when a contract fails that we tend to see Sentry logs?

All exceptions get reported to Sentry. Contract failures are not exceptions, so this PR makes sure we report these occurrences in Sentry even if they aren't exceptions.

arthurashman:
> > Is it only when a contract fails that we tend to see Sentry logs?
> 
> All exceptions get reported to Sentry. Contract failures are not exceptions, so this PR makes sure we report these occurrences in Sentry even if they aren't exceptions.

Makes sense, thank you

arthurashman (dismissed):

laurentqro:
> > > Is it only when a contract fails that we tend to see Sentry logs?
> > 
> > 
> > All exceptions get reported to Sentry. Contract failures are not exceptions, so this PR makes sure we report these occurrences in Sentry even if they aren't exceptions.
> 
> Makes sense, thank you

@arthurashman I turned contract failures into a custom Ruby Exception so we can dry up the code into the `ApplicationController` 👌🏻

chris-groves (approved):

arthurashman (approved):

PR #432, opened 2021-03-04
Add Sentry context for ActiveRecordInvalid errors

laurentqro:
## What

Currently, when a caseworker in VCD tries to link a defendant that is already marked as linked in CDA, the uniqueness constraint on `maat_reference` for linked `LaaReference` records means CDA returns a 500 error with no helpful error message (e.g. [this error](https://sentry.io/organizations/ministryofjustice/issues/2022856720)).

This PR ensures we:

* rescue from `ActiveRecord::RecordInvalid` and return hopefully helpful error messages in the response JSON
* add defendant_id and hearing_id to the extra context data that gets sent to Sentry in order to help with troubleshooting

## Checklist

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,11 +6,20 @@ class ApplicationController < ActionController::API
 
   ERROR_MAPPINGS = {
     ActionController::ParameterMissing => :bad_request,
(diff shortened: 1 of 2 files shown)


kmahern (approved):
lgtm 👍

PR #435, opened 2021-03-09
Add Sentry initializer

laurentqro:
Test Sentry message here: https://sentry.io/organizations/ministryofjustice/issues/2262883673/?environment=test&project=5375870&query=is%3Aunresolved&sort=date&statsPeriod=24h.

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,0 +1,6 @@
+if ENV["SENTRY_DSN"].present?
+  Sentry.init do |config|
+    config.dsn = ENV["SENTRY_DSN"]
+    config.breadcrumbs_logger = [:active_support_logger]
+  end
+end


config/initializers/sentry.rb
```diff
@@ -0,0 +1,4 @@
+Sentry.init do |config|
```
kmahern, line 1: Do you want some checks around this so that you are only logging Sentry errors if Rails.env is production? VCD has if Rails.env.eql?('production') && ENV['SENTRY_DSN'].present?

laurentqro: I want to log Sentry errors on all environments, so I might use only `ENV['SENTRY_DSN'].present?`

kmahern: I believe the Rails.env.eql?('production') check is to ensure sentry errors aren't raised when Rails.env is test or development, but as there isn't a SENTRY_DSN set it in that case it probably doesn't matter.

kmahern (approved):
👍

PR #436, opened 2021-03-09
WIP

arthurashman:
## What

## Checklist


app/services/hearings_creator.rb
```diff
@@ -51,17 +55,41 @@ def push_appeals
...
   end
 
-  def push_to_sqs(shared_time:, case_urn:, defendant:, appeal_data:)
+  def push_applications
+    hearing[:courtApplications]&.each do |application|
+      defendant_cases = application.dig(:applicant, :masterDefendant, :defendantCase)
+      return if defendant_cases.nil?
+
+      defendant_cases.each do |defendant_case|
+        defendant = LaaReference.find_by(defendant_id: defendant_case[:defendantId], linked: true)
```
laurentqro, line 70: Please make sure the variable name matches the thing that actually gets stored in it. When I see `defendant`, I expect a `Defendant` object of some kind to be in it.

app/services/sqs/publish_hearing.rb
```diff
@@ -138,12 +189,20 @@ def hearing_location(court_centre_id)
     end
 
     def session_hash
-      {
-        courtLocation: cjs_location,
-        dateOfHearing: defendant.dig(:offences, 0, :judicialResults, 0, :orderedDate),
-        postHearingCustody: PostHearingCustodyCalculator.call(offences: defendant[:offences]),
-        sessionValidateDate: defendant.dig(:offences, 0, :judicialResults, 0, :orderedDate),
-      }
+      if function_type == "OFFENCE"
```
laurentqro, line 192: I wonder if it wouldn't be better to post court applications to its own SQS queue. Then we wouldn't need this `function_type` field.

spec/services/hearings_creator_spec.rb
```diff
@@ -215,6 +250,105 @@
     end
   end
 
+  context "when an application to court" do
+    context "with a defendant that has a linked LAA reference" do
+      let(:prosecution_case_array) { nil }
+
+      it "calls the Sqs::PublishHearing service once" do
+        allow(LaaReference).to receive(:find_by).with(defendant_id: "dd22b110-7fbc-3036-a076-e4bb40d0a666", linked: true).and_return(defendant_one)
```
laurentqro, line 258: we're stubbing LaaReference to return a defendant here, which is not what LaaReference actually returns, so we're at risk of having our tests pass against our stubs but not against our actual code.

arthurashman: Ah, I've misunderstood something here then. The issue of expecting a Defendant object when a var is named defendant is stemming from the same misunderstanding. Will take another look. Thank you

app/services/sqs/publish_hearing.rb
arthurashman: Hmm makes sense, @msahilan07 would you be open to MAAT API subscribing to an applications to court queue so that it can differentiate that way?

spec/services/hearings_creator_spec.rb
```diff
@@ -215,6 +250,105 @@
...
+      it "does not call the Sqs::PublishHearing service" do
+        allow(LaaReference).to receive(:find_by).and_return(nil)
+
+        expect(Sqs::PublishHearing).not_to receive(:call)
+        create_hearings
+      end
+    end
+  end
+
+  context "when hearing had both a prosecution case and an applicationt to court" do
```
laurentqro, line 331: Might not be necessary to test this combination, as the two procedures are totally independent from one another.

app/services/hearings_creator.rb
```diff
@@ -51,17 +55,41 @@ def push_appeals
...
+        next if defendant.nil?
+
+        push_to_sqs(shared_time: shared_time,
+                    case_urn: application[:applicationReference],
```
laurentqro, line 74: After a talk with Ahilan, this should be accessed via the CDA database. But might be worth checking if it's present anywhere in the court_application data for us to pluck out.

spec/services/hearings_creator_spec.rb
laurentqro: It just returns an ActiveRecord object `LaaReference`.

app/services/hearings_creator.rb
arthurashman: I don't actually know what we want to send here to know what to look for in court_application

spec/services/hearings_creator_spec.rb
arthurashman: I thought it makes sense to test, because it is still the HearingCreator's responsibility to split the hearing into prosecution cases and court applications

PR #442, opened 2021-03-11
Clean up and self document dummy maat ref creation

laurentqro:
### Stop relying on little-used `dummy_maat_reference` boolean database column

This will remove the possibility of ending up with an inconsistent state between the maat reference field and and the `dummy_maat_reference`. Indeed, each one of these fields can be updated independently.

We should only ever determine whether a MAAT reference is a dummy or not by looking at the _actual_ maat reference.

### Refactorings

* move all dummy reference logic to the `LaaReference` model
* use more intention-revealing method names

## Note

There is a PR to drop the `dummy_reference` boolean column from `laa_references` table, branched off of the present one: https://github.com/ministryofjustice/laa-court-data-adaptor/pull/443.

## Checklist

### DIFF
--- a/app/models/laa_reference.rb
+++ b/app/models/laa_reference.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class LaaReference < ApplicationRecord
+  self.ignored_columns = %w[dummy_maat_reference]
+
   validates :defendant_id, presence: true
   validates :maat_reference, presence: true, uniqueness: { conditions: -> { where(linked: true) } }
   validates :user_name, presence: true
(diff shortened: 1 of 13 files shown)


app/models/laa_reference.rb
```diff
@@ -1,7 +1,25 @@
 # frozen_string_literal: true
 
 class LaaReference < ApplicationRecord
+  self.ignored_columns = %w[dummy_maat_reference]
```
laurentqro, line 4: First step of two in dropping the `dummy_maat_reference` column

https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-ignored_columns-3D

kmahern:
Looks good from my perspective and I do think it makes the code clearer and it is the right thing to remove the reliance on the dummy_maat_reference boolean database column. I don't feel I am familiar enough with the CDA codebase to review this fully though - is someone from the CDA team available? I would also suggest it is probably a change we should do some manual testing around.

arthurashman (approved):
This is a really lovely refactor! Thank you!

PR #447, opened 2021-03-12
Refactorings on HearingsCreator

laurentqro:
☝🏻

### DIFF
--- a/spec/services/hearings_creator_spec.rb
+++ b/spec/services/hearings_creator_spec.rb
@@ -93,10 +93,13 @@ RSpec.describe HearingsCreator do
         },
         "prosecutionCases": prosecution_case_array,
         "courtApplications": court_applications_array,
+        "hearingDays": hearing_days,
       },
       "sharedTime": "2018-10-25 11:30:00",
     }
   end
+
+  let(:hearing_days) { [{ sittingDay: "2021-02-12" }] }
 
   let(:hearing) { Hearing.create!(body: hearing_body) }
 
@@ -113,7 +116,7 @@ RSpec.describe HearingsCreator do
                                                                                hearing_data: {
                                                                                  jurisdiction_type: "MAGISTRATES",
                                                                                  court_centre_id: "dd22b110-7fbc-3036-a076-e4bb40d0a519",
-                                                                                 court_centre_code: "123",
+                                                                                 hearing_days: hearing_days,
                                                                                }))
         create_hearings
       end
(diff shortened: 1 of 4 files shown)


spec/services/hearings_creator_spec.rb
```diff
@@ -250,56 +250,51 @@
...
-    context "with two defendants one with linked LAA reference, one without" do
+    context "with two defendants, one with linked LAA reference, one without" do
       let(:prosecution_case_array) { nil }
       let(:defendant_case_array) { [defendant_case_one, defendant_case_two] }
 
-      it "calls the Sqs::PublishHearing service twice" do
-        allow(LaaReference).to receive(:find_by).with(defendant_id: "dd22b110-7fbc-3036-a076-e4bb40d0a666", linked: true).and_return(defendant_one)
-        allow(LaaReference).to receive(:find_by).with(defendant_id: "ad22b110-7fbc-3036-a076-e4bb40d0a667", linked: true).and_return(nil)
+      it "calls the Sqs::PublishHearing service once" do
+        LaaReference.create!(defendant_id: "dd22b110-7fbc-3036-a076-e4bb40d0a666", linked: true, maat_reference: "123", user_name: "Bob")
```
arthurashman, line 291: So this creates a real LaaReference and puts it into the db. Is there any way to set up the test without doing this? Like stubbing `LaaReference.find_by` and setting it to return the boolean we want for the test. I don't like the calling of the real thing in the tests when it's then depending on a second class working properly

spec/services/hearings_creator_spec.rb
```diff
@@ -328,27 +323,6 @@
...
-                                                                               court_centre_id: "dd22b110-7fbc-3036-a076-e4bb40d0a519"))
-        expect(Sqs::PublishHearing).to receive(:call).once.with(hash_including(shared_time: "2018-10-25 11:30:00",
-                                                                               jurisdiction_type: "MAGISTRATES",
-                                                                               case_urn: "12345",
-                                                                               defendant: defendant_one,
-                                                                               court_centre_id: "dd22b110-7fbc-3036-a076-e4bb40d0a519"))
-        create_hearings
-      end
-    end
-  end
```
arthurashman, line 350: Do you not think this is worth checking? That it handles a hearing with one of each correctly?

arthurashman (approved):

spec/services/hearings_creator_spec.rb
laurentqro: Sorry, what don't you like about it?

laurentqro:
superseded by https://github.com/ministryofjustice/laa-court-data-adaptor/pull/449

PR #453, opened 2021-03-17
Remove delay between HTTP requests

laurentqro:
## What

The initial intention was to introduce some rate limiting because we suspected CP could not handle it. CP should be able to handle this load.

The Faraday logger middleware introduced in https://github.com/ministryofjustice/laa-court-data-adaptor/pull/451 means we can monitor responses from CP. Removing the delay allows us to see any HTTP response statuses that could indicate potential issues on CP's side.

### DIFF
--- a/app/services/api/record_laa_reference.rb
+++ b/app/services/api/record_laa_reference.rb
@@ -2,8 +2,6 @@
 
 module Api
   class RecordLaaReference < ApplicationService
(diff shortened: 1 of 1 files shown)


arthurashman (approved):

laurentqro:
PR https://github.com/ministryofjustice/laa-court-data-adaptor/pull/453 further justifies the removal of the `sleep`s.

PR #460, opened 2021-03-19
Handle HearingEventRecording with blank body

laurentqro:
This PR fixes and edge case where a hearing log request to Common Platform returns `{}` in the response body, even though the schema forbids it.

This fix is put in place until Common Platform make a fix to their API response.

### DIFF
--- a/spec/models/hearing_spec.rb
+++ b/spec/models/hearing_spec.rb
@@ -55,6 +55,12 @@ RSpec.describe Hearing, type: :model do
 
           it { expect(hearing.hearing_events).to all be_a(HearingEvent) }
         end
+
+        context "with a missing events key from hearing log response" do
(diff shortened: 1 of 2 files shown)


msahilan07 (approved):

spec/models/hearing_spec.rb
```diff
@@ -53,7 +53,15 @@
         context "with blank hearing events" do
           let(:hearing_events) { [hearing_event_recording, nil] }
 
-          it { expect(hearing.hearing_events).to all be_a(HearingEvent) }
+          it do
```
arthurashman, line 56: Is there a reason this changed? Does it need a test name?

arthurashman (dismissed):

msahilan07 (approved):

arthurashman (approved):

PR #467, opened 2021-03-23
Configure Faraday to retry for 429 Too Many Requests HTTP responses

laurentqro:
Common Platform currently enforces a 10 requests per minute rate limit. To make sure no request gets lost, this PR configures the Faraday HTTP client to automatically retry after the amount of time specified in the `Retry-After` response header.

Faraday documentation: https://github.com/lostisland/faraday/blob/master/docs/middleware/request/retry.md#automatically-handle-the-retry-after-header


msahilan07 (approved):
LGTM

PR #479, opened 2021-03-29
Remove Crown Court Outcome logic

laurentqro:
## What

Crown Court Outcomes are computed by MAAT API after receiving prosecution case data from CDA. CDA’s role is only to pass on the data to MAAT API so it can compute these outcomes. This logic should therefore be removed from CDA.

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-549)


arthurashman (approved):
This looks good to me, but mostly because I was convinced push appeals wasn't doing anything anyway. If that assumption is wrong though, and I was only looking at it from a new schema perspective, we will need to wait until we're confirmed receiving CP data per the new schema before removing this.

arthurashman (approved):

laurentqro:
We need to go back to the drawing board, we don't want to remove all the Crown Court outcome logic at this point.

PR #481, opened 2021-03-30
Serialize hearings with no prosecution cases

laurentqro:
The current `HearingSerializer` errors with `undefined method `flat_map' for nil:NilClass` because some hearings can have no prosecution cases. It's an optional field as per HMCTS schema. This currently causes a 500 server error on VCD.

This PR:

* makes sure we are able to serialize a hearing object with no associated prosecution cases,
* serializes empty collections as `[]` rather than `null` by casting `nil` into `[]` with `Array(nil)`.

Fixes issue https://github.com/ministryofjustice/laa-court-data-adaptor/issues/480.

### DIFF
--- a/app/models/hearing.rb
+++ b/app/models/hearing.rb
@@ -12,7 +12,7 @@ class Hearing < ApplicationRecord
   end
 
   def hearing_days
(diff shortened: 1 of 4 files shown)


msahilan07 (approved):
lgtm

Issue #188, opened 2020-06-12
RESTful routes for linking and unlinking


jsugarman:
**What**
The ability to "unlink" via a destroy action on the laa_references controller

**Why**
"View court data" has duplicated the logic of linking via the [laa_reference create endpoint](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/controllers/api/internal/v1/laa_references_controller.rb) and unlinking via the [defendant update endpoint](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/master/app/controllers/api/internal/v1/defendants_controller.rb). It would be less surprising/more conventional if these were RESTful actions on the same endpoint - i.e. so `laa_references: create`  for linking, `laa_references: destroy`  for unlinking.

PR #483, opened 2021-03-31
Create RESTful route for unlinking

laurentqro:
## What

The ability to "unlink" via a destroy action on the `laa_references` controller.

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-568)

Addresses https://github.com/ministryofjustice/laa-court-data-adaptor/issues/188.

## Why

VCD has duplicated logic for linking via the `laa_reference` create endpoint and unlinking via the defendant update endpoint.

This PR makes it less surprising/more conventional by making linking and unlinking RESTful actions on the same endpoint.

### DIFF
--- a/app/controllers/api/internal/v1/laa_references_controller.rb
+++ b/app/controllers/api/internal/v1/laa_references_controller.rb
@@ -5,29 +5,42 @@ module Api
     module V1
       class LaaReferencesController < ApplicationController
         def create
-          enforce_contract!
-          create_laa_reference
+          contract = NewLaaReferenceContract.new.call(**transformed_params)
+          enforce_contract!(contract)
+
+          LaaReferenceCreator.call(defendant_id: transformed_params[:defendant_id],
+                                   user_name: transformed_params[:user_name],
+                                   maat_reference: transformed_params[:maat_reference])
+
+          render status: :accepted
+        end
+
+        def destroy
+          contract = UnlinkDefendantContract.new.call(**transformed_params)
+          enforce_contract!(contract)
+
+          UnlinkLaaReferenceWorker.perform_async(
+            Current.request_id,
+            transformed_params[:defendant_id],
+            transformed_params[:user_name],
+            transformed_params[:unlink_reason_code],
+            transformed_params[:unlink_other_reason_text],
+          )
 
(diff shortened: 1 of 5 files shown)


app/controllers/api/internal/v1/laa_references_controller.rb
```diff
@@ -5,41 +5,56 @@ module Internal
...
           render status: :accepted
         end
 
       private
 
-        def enforce_contract!
+        def enforce_contract!(contract)
           unless contract.success?
-            message = "LaaReference contract failed with: #{contract.errors.to_hash}"
+            message = "Contract failed with: #{contract.errors.to_hash}"
```
arthurashman, line 37: Will contract.errors be verbose enough for us to know if it is the linking or unlinking that has failed? Or does sentry literally point to this file/line of code anyway?

spec/requests/api/internal/v1/laa_references_spec.rb
```diff
@@ -154,4 +156,62 @@
...
+        parameter "$ref" => "#/components/parameters/transaction_id_header"
+
+        before do
+          expect(UnlinkLaaReferenceWorker).not_to receive(:perform_async)
+        end
+
+        run_test!
+      end
+    end
+  end
```
arthurashman, line 216: Do you know of any good resources that explain how to write these tests/how they work?

arthurashman (approved):

spec/requests/api/internal/v1/laa_references_spec.rb
laurentqro: Sure:

https://github.com/rswag/rswag and https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec.

app/controllers/api/internal/v1/laa_references_controller.rb
laurentqro: Good question. Sentry will tell us which controller action the error is happening on, so we should have enough information for debugging.

e.g. https://sentry.io/organizations/ministryofjustice/issues/2266587125 `Errors::ContractError Api::Internal::V1::LaaReferencesController#create`.

PR #488, opened 2021-04-06
Refactor cracked ineffective trial model/serializer specs

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-575)

For each serialiser in CDA, test that it can serialise a resource when it has only the required fields, and when it has all the fields.

## Why

CDA frequently sees an entire class of bugs that are due to the fact that our serialisers assume an optional field is always present.

e.g. error https://sentry.io/organizations/ministryofjustice/issues/2303812534/?environment=uat&project=5375870&query=is%3Aresolved&statsPeriod=14d.


spec/models/cracked_ineffective_trial_spec.rb
```diff
@@ -1,35 +1,16 @@
...
-  end
-
-  describe "#type" do
-    subject(:type) { cracked_ineffective_trial.type }
-
-    it "is expected to downcase the type" do
-      expect(type).to eql("cracked")
+  context "with required fields" do
+    let(:cracked_ineffective_trial_data) do
+      JSON.parse(file_fixture("cracked_ineffective_trial/required_fields.json").read)
```
arthurashman, line 8: So there is an optional field of `data` - do you want to do a context with all fields and another file_fixture? Or is that unnecessary/overkill?

arthurashman (approved):

PR #493, opened 2021-04-08
Test defence organisation model with required and optional fields

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-575)

For each serialiser in CDA, test that it can serialise a resource when it has only the required fields, and when it has all the fields.

## Why

CDA frequently sees an entire class of bugs that are due to the fact that our serialisers assume an optional field is always present.


spec/models/defence_organisation_spec.rb
```diff
@@ -1,18 +1,16 @@
 # frozen_string_literal: true
 
 RSpec.describe DefenceOrganisation, type: :model do
-  subject(:defendant) { described_class.new(body: defence_organisation_hash) }
-
-  let(:defence_organisation_hash) do
-    JSON.parse(file_fixture("defence_organisation.json").read)
+  subject(:defendant) do
```
arthurashman, line 4: Would this be clearer as `defence_organisation`?

spec/serializer/defence_organisation_serializer_spec.rb
```diff
@@ -3,27 +3,32 @@
...
-                    name: "The Johnson Partnership",
-                    address1: "104",
-                    address2: "Fleet Street",
-                    address3: "Westminster",
-                    address4: "London",
-                    address5: "GB",
-                    postcode: "EC4A 2AH")
+  context "with required fields" do
+    let(:defence_organisation_data) { JSON.parse(file_fixture("defence_organisation/required_fields.json").read) }
+    let(:defence_organisation) { DefenceOrganisation.new(body: defence_organisation_data) }
```
arthurashman, line 8: My preference would be to have a fake DefenceOrganisation here so we are testing the serializer in isolation of the model. What are your thoughts?

spec/serializer/defence_organisation_serializer_spec.rb
```diff
@@ -3,27 +3,32 @@
...
+    it { expect(attributes[:address4]).to be_nil }
+    it { expect(attributes[:address5]).to be_nil }
+    it { expect(attributes[:postcode]).to be_nil }
   end
 
-  context "with attributes" do
-    let(:attribute_hash) { subject[:data][:attributes] }
+  context "with all fields" do
+    let(:defence_organisation_data) { JSON.parse(file_fixture("defence_organisation/all_fields.json").read) }
+    let(:defence_organisation) { DefenceOrganisation.new(body: defence_organisation_data) }
```
arthurashman, line 23: Same here obviously

arthurashman (dismissed):

spec/serializer/defence_organisation_serializer_spec.rb
laurentqro: Yeah good point. And then we can use the fixtures to test the model itself.

spec/serializer/defence_organisation_serializer_spec.rb
```diff
@@ -15,15 +15,13 @@
                     postcode: "EC4A 2AH")
   end
 
-  context "with attributes" do
-    let(:attribute_hash) { subject[:data][:attributes] }
+  let(:attributes) { subject[:data][:attributes] }
```
arthurashman, line 19: Sorry to be a pedant, but just for completion, could we have 
`it { expect(attributes[:id]).to eq("UUID") }`

Or even 11223344 if you want to change the double too

spec/models/defence_organisation_spec.rb
```diff
@@ -1,18 +1,29 @@
...
-  it { expect(defendant.name).to eq("SOME ORGANISATION") }
-  it { expect(defendant.address1).to eq("102") }
-  it { expect(defendant.address2).to eq("Petty France") }
-  it { expect(defendant.address3).to eq("Floor 5") }
-  it { expect(defendant.address4).to eq("St James") }
-  it { expect(defendant.address5).to eq("Westminster") }
-  it { expect(defendant.postcode).to eq("EC4A 2AH") }
+  context "with all fields" do
+    let(:defence_organisation_data) { JSON.parse(file_fixture("defence_organisation/all_fields.json").read) }
+
```
arthurashman, line 20: `it { expect(defence_organisation.name.id).to eq("11223344") }`

spec/models/defence_organisation_spec.rb
```diff
@@ -1,18 +1,29 @@
...
 
 RSpec.describe DefenceOrganisation, type: :model do
-  subject(:defendant) { described_class.new(body: defence_organisation_hash) }
+  subject(:defence_organisation) { described_class.new(body: defence_organisation_data) }
 
-  let(:defence_organisation_hash) do
-    JSON.parse(file_fixture("defence_organisation.json").read)
+  context "with required fields" do
+    let(:defence_organisation_data) { JSON.parse(file_fixture("defence_organisation/required_fields.json").read) }
+
```
arthurashman, line 8: `it { expect(defence_organisation.name.id).to be_nil }`

arthurashman (changes requested):

spec/models/defence_organisation_spec.rb
laurentqro: another good spot, thanks!

arthurashman (approved):

PR #500, opened 2021-04-14
refactor push prosecution case to use new MessagePublisher class

arthurashman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-LASB570)

Refactors to reuse MessagePublisher class, and allow for CDA to split apart Hearing payloads from Common Platform and send different messages to MAATAPI SQS queue dependent on whether it is prosecution case data or application to court data. 

## Checklist


spec/models/maat_api/prosecution_case_spec.rb
```diff
@@ -0,0 +1,247 @@
...
+    it "has a case URN" do
+      expect(prosecution_case.case_urn).to eql("20GD0217100")
+    end
+
+    it "has a jurisdiction type" do
+      expect(prosecution_case.jurisdiction_type).to eql("CROWN")
+    end
+
+    it "has a defendant ASN" do
+      expect(prosecution_case.defendant_asn).to be_nil
```
laurentqro, line 161: `has no`

app/models/hmcts_common_platform/offence.rb
```diff
@@ -0,0 +1,75 @@
...
+    def laa_appln_reference_status_description
+      data.dig(:laaApplnReference, :statusDescription)
+    end
+
+    def laa_appln_reference_laa_contract_number
+      data.dig(:laaApplnReference, :laaContractNumber)
+    end
+
+    def results
+      data[:judicialResults]&.map do |judicial_result_data|
```
laurentqro, line 62: `Array(data[:judicialResults]).map...`

app/models/maat_api/prosecution_case.rb
```diff
@@ -0,0 +1,175 @@
...
+    def case_creation_date
+      @shared_time.to_date.strftime("%Y-%m-%d")
+    end
+
+    def cjs_location
+      court_centre.short_oucode
+    end
+
+    def doc_language
+      # Should we derive this from somewhere else?
```
laurentqro, line 33: Indeed, it's worth asking Ahilan if we want to get this with `person_defendant.documentation_language_needs` like we do for Court Applications. Let's make a note of it and do it in a separate PR.

app/services/hearings_creator.rb
```diff
@@ -20,14 +20,24 @@ def call
...
-      prosecution_case[:defendants].each do |defendant|
-        next if defendant[:offences].map { |offence| offence.dig(:laaApplnReference, :applicationReference)&.start_with?("A", "Z") }.any?
-
-        push_to_sqs(shared_time: shared_time,
-                    case_urn: prosecution_case[:prosecutionCaseIdentifier][:caseURN],
-                    defendant: defendant,
-                    appeal_data: nil)
+    hearing[:prosecutionCases]&.each do |prosecution_case_data|
+      prosecution_case_data[:defendants].each do |defendant_data|
+        next if defendant_data[:offences].map { |offence| offence.dig(:laaApplnReference, :applicationReference)&.start_with?("A", "Z") }.any?
```
laurentqro, line 25: slightly shorter alternative: `next if defendant_data[:offences].any? { |offence| offence.dig(:laaApplnReference, :applicationReference)&.start_with?("A", "Z") }`

app/models/hmcts_common_platform/offence.rb
arthurashman: This then returns an empty array if there are no judicialResults which is not the output we want and would then need to be changed to nil at another point

app/services/hearings_creator.rb
```diff
@@ -20,14 +20,24 @@ def call
...
-        push_to_sqs(shared_time: shared_time,
-                    case_urn: prosecution_case[:prosecutionCaseIdentifier][:caseURN],
-                    defendant: defendant,
-                    appeal_data: nil)
+    hearing[:prosecutionCases]&.each do |prosecution_case_data|
+      prosecution_case_data[:defendants].each do |defendant_data|
+        next if defendant_data[:offences].any? { |offence| offence.dig(:laaApplnReference, :applicationReference)&.start_with?("A", "Z") }
+
+        laa_reference = LaaReference.find_by!(defendant_id: defendant_data[:id], linked: true)
+        next if laa_reference.blank?
```
laurentqro, line 28: I don't think we need this. It wasn't in the original implementation, and we now know that all incoming cases are linked.

arthurashman: Quite right. Thank you

chris-groves (approved):

spec/models/maat_api/prosecution_case_spec.rb
```diff
@@ -43,11 +43,16 @@
   end
 
   it "has no crown_court_outcome" do
```
laurentqro, line 45: `it has a crown_court_outcome` :)

spec/models/maat_api/prosecution_case_spec.rb
```diff
@@ -43,11 +43,16 @@
...
+      ccooOutcome: "CONVICTED",
+      caseEndDate: "2021-04-10",
+      appealType: nil,
+    }
+    expect(prosecution_case.crown_court_outcome).to eql(expected)
   end
 
   it "has an inactive" do
-    expect(prosecution_case.inactive).to eql("N")
+    expect(prosecution_case.inactive).to eql("Y")
```
laurentqro, line 55: why the change here?

spec/models/maat_api/prosecution_case_spec.rb
```diff
@@ -15,7 +15,7 @@
   end
 
   it "has a jurisdiction type" do
-    expect(prosecution_case.jurisdiction_type).to eql("MAGISTRATES")
+    expect(prosecution_case.jurisdiction_type).to eql("CROWN")
```
laurentqro, line 18: why the change here?

laurentqro: ok I get it now :)

spec/models/hmcts_common_platform/verdict_spec.rb
```diff
@@ -0,0 +1,88 @@
+RSpec.describe HmctsCommonPlatform::Verdict, type: :model do
+  let(:data) { JSON.parse(file_fixture("verdict/all_fields.json").read).deep_symbolize_keys }
```
laurentqro, line 2: should we put this in a context block?

spec/models/hmcts_common_platform/defendant_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  describe "offences" do
+    it "are HmctsCommonPlatform::Offence objects" do
+      expect(defendant.offences).to all(be_a(HmctsCommonPlatform::Offence))
+      expect(defendant.offences).to be_a(Array)
+    end
+  end
+end
```
laurentqro, line 85: add a test case for required fields only?

laurentqro: I've got this you can use: https://github.com/ministryofjustice/laa-court-data-adaptor/pull/494/files#diff-d1d27e19030c31cfb480424ff6c4cf675d01f5852ca9897c4b57dec2744f637fR1

spec/models/maat_api/prosecution_case_spec.rb
arthurashman: Because I have set the default fixture file to being "CROWN" jurisdiction (for CCO) and so inactive calculates to "Y"

chris-groves (dismissed):

laurentqro (approved):
great work

PR #514, opened 2021-04-30
Use HashWithIndifferentAccess for storing HMCTS raw data

laurentqro:
## What

Use HashWithIndifferentAccess for storing HMCTS raw data, so as to avoid not being able to access the hash key's value because we're accessing with symbolised rather than stringified keys.

Documentation: https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html

We also split `hearing/all_fields.json` JSON fixture file into `hearing_resulted.json` and removed the `"sharedTime"` and `"hearing"` keys from `hearing/all_fields.json`, which are two distinct entities as per the schema, and are not be confused with one another.

### DIFF
--- a/app/models/hmcts_common_platform/court_application.rb
+++ b/app/models/hmcts_common_platform/court_application.rb
@@ -2,11 +2,11 @@ module HmctsCommonPlatform
   class CourtApplication
     attr_reader :data
 
-    def initialize(data)
-      @data = data || {}
-    end
-
     delegate :blank?, to: :data
+
+    def initialize(data)
+      @data = HashWithIndifferentAccess.new(data || {})
+    end
 
     def type_id
       data.dig(:type, :id)
--- a/app/models/hmcts_common_platform/delegated_powers.rb
+++ b/app/models/hmcts_common_platform/delegated_powers.rb
@@ -2,11 +2,11 @@ module HmctsCommonPlatform
   class DelegatedPowers
     attr_reader :data
 
+    delegate :blank?, to: :data
+
     def initialize(data)
       @data = HashWithIndifferentAccess.new(data || {})
     end
-
-    delegate :blank?, to: :data
 
     def user_id
       data[:userId]
--- a/app/models/hmcts_common_platform/delegated_powers.rb
+++ b/app/models/hmcts_common_platform/delegated_powers.rb
@@ -3,7 +3,7 @@ module HmctsCommonPlatform
     attr_reader :data
 
     def initialize(data)
-      @data = data || {}
+      @data = HashWithIndifferentAccess.new(data || {})
     end
 
     delegate :blank?, to: :data
--- a/app/models/hmcts_common_platform/offence.rb
+++ b/app/models/hmcts_common_platform/offence.rb
@@ -2,11 +2,11 @@ module HmctsCommonPlatform
   class Offence
     attr_reader :data
 
-    def initialize(data)
-      @data = data || {}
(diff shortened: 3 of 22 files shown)


app/models/hmcts_common_platform/court_application.rb
```diff
@@ -2,12 +2,12 @@ module HmctsCommonPlatform
   class CourtApplication
     attr_reader :data
 
+    delegate :blank?, to: :data
+
     def initialize(data)
       @data = data || {}
```
arthurashman, line 8: Any reason this isn't `@data = HashWithIndifferentAccess.new(data || {})` ?

app/models/hmcts_common_platform/delegated_powers.rb
```diff
@@ -3,7 +3,7 @@ class DelegatedPowers
     attr_reader :data
 
     def initialize(data)
-      @data = data || {}
+      @data = HashWithIndifferentAccess.new(data || {})
     end
 
     delegate :blank?, to: :data
```
arthurashman, line 9: Move this above initialize?

app/models/hmcts_common_platform/offence.rb
```diff
@@ -3,7 +3,7 @@ class Offence
     attr_reader :data
 
     def initialize(data)
-      @data = data || {}
+      @data = HashWithIndifferentAccess.new(data || {})
     end
 
     delegate :blank?, to: :data
```
arthurashman, line 9: Move this above initialize?

app/models/hmcts_common_platform/person_defendant.rb
```diff
@@ -3,7 +3,7 @@ class PersonDefendant
     attr_reader :data
 
     def initialize(data)
-      @data = data || {}
+      @data = HashWithIndifferentAccess.new(data || {})
     end
 
     delegate :blank?, to: :data
```
arthurashman, line 9: Move this above initialize?

app/models/hmcts_common_platform/plea.rb
```diff
@@ -3,7 +3,7 @@ class Plea
     attr_reader :data
 
     def initialize(data)
-      @data = data || {}
+      @data = HashWithIndifferentAccess.new(data || {})
     end
 
     delegate :blank?, to: :data
```
arthurashman, line 9: Move this above initialize?

spec/models/hmcts_common_platform/defendant_spec.rb
```diff
@@ -2,7 +2,7 @@
   let(:defendant) { described_class.new(data) }
 
   context "when the defendant has all fields" do
-    let(:data) { JSON.parse(file_fixture("defendant/all_fields.json").read).deep_symbolize_keys }
+    let(:data) { JSON.parse(file_fixture("defendant/all_fields.json").read) }
```
arthurashman, line 5: Out of curiosity, do the serialisers in the app sybolize?

app/models/hmcts_common_platform/court_application.rb
laurentqro: there is a bad reason: I forgot this one :)

app/models/hmcts_common_platform/delegated_powers.rb
arthurashman: Last one 😉

arthurashman (approved):

PR #533, opened 2021-05-12
Create / update data-wrapper objects

laurentqro:
Create / update data-wrapper objects for us to work with across the application.

### DIFF
--- a/app/models/hmcts_common_platform/court_application.rb
+++ b/app/models/hmcts_common_platform/court_application.rb
@@ -4,28 +4,14 @@ module HmctsCommonPlatform
 
     delegate :blank?, to: :data
 
+    delegate :id, :description, :code, :category_code, :legislation, to: :court_application_type, prefix: true
+
     def initialize(data)
       @data = HashWithIndifferentAccess.new(data || {})
     end
 
-    def type_id
-      data.dig(:type, :id)
-    end
-
-    def type_name
-      data.dig(:type, :type)
-    end
-
-    def type_code
-      data.dig(:type, :code)
-    end
-
-    def type_category_code
-      data.dig(:type, :categoryCode)
-    end
-
-    def type_legislation
-      data.dig(:type, :legislation)
+    def id
+      data[:id]
     end
 
     def application_particulars
@@ -104,6 +90,12 @@ module HmctsCommonPlatform
       person_defendant.email_secondary
     end
 
+    def respondents
+      Array(data[:respondents]).map do |respondent_data|
+        HmctsCommonPlatform::CourtApplicationParty.new(respondent_data)
+      end
+    end
+
     def judicial_results
       data[:judicialResults]&.map do |judicial_result_data|
         HmctsCommonPlatform::JudicialResult.new(judicial_result_data)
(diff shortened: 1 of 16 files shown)


spec/models/hmcts_common_platform/court_application_spec.rb
```diff
@@ -100,6 +100,12 @@
     end
   end
 
+  describe "respondents" do
+    it "are HmctsCommonPlatform::CourtApplicationParty objects" do
+      expect(court_application.respondents).to all(be_a(HmctsCommonPlatform::CourtApplicationParty))
+    end
+  end
+
```
arthurashman, line 108: No tests for `judicial_result_ids` or `respondent_ids`?

arthurashman: Or `court_application_type` ?

Is this one only used locally? Worth making it private?

app/models/hmcts_common_platform/court_application.rb
```diff
@@ -104,6 +90,24 @@ def defendant_email_secondary
       person_defendant.email_secondary
     end
 
+    def court_application_type
```
arthurashman, line 93: Is this one only used locally? Worth making it private?

spec/models/hmcts_common_platform/judicial_result_spec.rb
```diff
@@ -6,8 +6,8 @@
     expect(data).to match_json_schema(:judicial_result)
   end
 
-  it "has a code" do
-    expect(judicial_result.code).to eql("4600")
+  it "has a CJS code" do
```
arthurashman, line 9: Can we add a test for `id` too?

app/models/hmcts_common_platform/court_application_type.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def initialize(data)
+      @data = HashWithIndifferentAccess.new(data || {})
+    end
+
+    def id
+      data[:id]
+    end
+
+    def description
```
arthurashman, line 13: Why not type? just avoiding _type.type?

app/models/hmcts_common_platform/court_application.rb
```diff
@@ -4,28 +4,14 @@ class CourtApplication
 
     delegate :blank?, to: :data
 
+    delegate :id, :description, :code, :category_code, :legislation, to: :court_application_type, prefix: true
```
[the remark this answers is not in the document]
laurentqro: me neither, it was Rubocop that introduced me to this! :)

spec/models/hmcts_common_platform/court_application_spec.rb
laurentqro: yeah good point, worth making `court_application_type` private.

laurentqro: I'm going to remove `judicial_result_ids` and `respondent_ids` from this PR, it's outside of the scope. I'd also like to see if I can keep these out of our objects, because they're required only by the JSON API Serializer gem.

app/models/hmcts_common_platform/court_application_type.rb
laurentqro: yeah, not to mention `type` is a super generic word that also has a meaning in programming, so it's a good one to avoid.

arthurashman (approved):

PR #543, opened 2021-05-26
Ensure nil judicial results processed as empty array

arthurashman:
## What

Judicial results is not a required field on Court Applications. This makes sure if null, judicial results is still processed as an Array and avoids NoMethodError for NilClass 

## Checklist


laurentqro (approved):
Brilliant, thank you, I stumbled upon this error earlier!

PR #545, opened 2021-05-27
Persist LaaReference after posting to CP

laurentqro:
## What

Persist LaaReference in the CDA database only after posting to Common Platform.

## Why

Prevent CDA from getting out of sync with Common Plaform.


app/services/maat_link_creator.rb
```diff
@@ -1,13 +1,20 @@
...
+      defendant_id: defendant_id,
+      user_name: user_name,
+      maat_reference: maat_reference.presence || LaaReference.generate_linking_dummy_maat_reference,
+    )
   end
 
   def call
     publish_laa_reference_to_queue unless laa_reference.dummy_maat_reference?
     post_laa_references_to_common_platform
+    persist_laa_reference!
```
arthurashman, line 17: So if `post_laa_references_to_common_platform` fails, do we definitely not hit this line?

If response from RecordLaaReference.call is status 500, is the default behaviour to throw an error?

app/services/laa_reference_unlinker.rb
```diff
@@ -10,9 +10,9 @@ def initialize(defendant_id:, user_name:, unlink_reason_code:, unlink_other_reas
   end
 
   def call
-    unlink_maat_reference!
     push_to_queue unless laa_reference.dummy_maat_reference?
     update_offences_on_common_platform
```
arthurashman, line 14: Likewise here, will the code stop before reaching the unlink method if CP fails?

arthurashman (approved):

app/controllers/api/internal/v1/laa_references_controller.rb
```diff
@@ -8,9 +8,12 @@ def create
           contract = NewLaaReferenceContract.new.call(**transformed_params)
           enforce_contract!(contract)
 
-          LaaReferenceCreator.call(defendant_id: transformed_params[:defendant_id],
-                                   user_name: transformed_params[:user_name],
-                                   maat_reference: transformed_params[:maat_reference])
+          MaatLinkCreatorWorker.perform_async(
```
arthurashman, line 11: Is there a reason these aren't named params?

arthurashman (approved):

app/controllers/api/internal/v1/laa_references_controller.rb
laurentqro: Tried with named params, but doesn't work.

It's the way Sidekiq's `perform_async` method is implemented: https://github.com/mperham/sidekiq/blob/3330df0ee37cfd3e0cd3ef01e3e66b584b99d488/lib/sidekiq/worker.rb#L163

arthurashman (approved):

PR #558, opened 2021-06-09
Create data wrappers for ProsecutionCase and LaaReference

laurentqro:
## What

* Create data wrappers for ProsecutionCase and LaaReference
* Add methods to access data on various other types

### DIFF
--- a/app/models/hmcts_common_platform/offence.rb
+++ b/app/models/hmcts_common_platform/offence.rb
@@ -4,6 +4,8 @@ module HmctsCommonPlatform
 
     delegate :blank?, to: :data
 
+    delegate :application_reference, :status_code, :status_date, :status_description, :laa_contract_number, :effective_end_date, to: :laa_reference, prefix: true
+
     def initialize(data)
       @data = HashWithIndifferentAccess.new(data || {})
     end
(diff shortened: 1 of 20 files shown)


app/models/hmcts_common_platform/offence.rb
```diff
@@ -40,24 +42,8 @@ def allocation_decision_mot_reason_code
...
-      data.dig(:laaApplnReference, :effectiveEndDate)
-    end
-
-    def laa_appln_reference_status_description
-      data.dig(:laaApplnReference, :statusDescription)
-    end
-
-    def laa_appln_reference_laa_contract_number
-      data.dig(:laaApplnReference, :laaContractNumber)
+    def laa_reference
```
arthurashman, line 45: Thinking this can be private. In the same way `person_defendant` is private in `Defendant`

spec/models/hmcts_common_platform/offence_spec.rb
```diff
@@ -69,9 +69,13 @@
       expect(offence.plea).to be_a(HmctsCommonPlatform::Plea)
     end
 
-    it "has a verdict offence id" do
+    it "has a verdict" do
       expect(offence.verdict).to be_a(HmctsCommonPlatform::Verdict)
     end
+
+    it "has an LAA reference" do
```
arthurashman, line 76: idk if this is ever called like this - if not can we remove this test and make the method private?

arthurashman (approved):

PR #561, opened 2021-06-10
Refactor HearingsCreator to use data-wrapper objects only

laurentqro:
## What

* Rename body to hearing_resulted_data
* Create extra data-wrapper objects
* Refactor HearingsCreator to use data-wrapper objects only
* Refactor MaatApi::CourtApplication
* MaatApi::ProsecutionCase

### DIFF
--- a/app/models/maat_api/court_application.rb
+++ b/app/models/maat_api/court_application.rb
@@ -1,13 +1,10 @@
 module MaatApi
   class CourtApplication
-    attr_reader :hearing, :court_application, :maat_reference
+    attr_reader :hearing_resulted, :court_application, :maat_reference
 
-    delegate :jurisdiction_type, to: :hearing
-
-    def initialize(hearing_body, court_application_data, maat_reference)
-      @shared_time = hearing_body[:sharedTime]
-      @hearing = HmctsCommonPlatform::Hearing.new(hearing_body[:hearing])
-      @court_application = HmctsCommonPlatform::CourtApplication.new(court_application_data)
+    def initialize(hearing_resulted, court_application, maat_reference)
+      @hearing_resulted = hearing_resulted
+      @court_application = court_application
       @maat_reference = maat_reference
     end
 
@@ -18,7 +15,7 @@ module MaatApi
     end
 
     def cjs_area_code
-      find_court_centre_by_id(hearing.court_centre_id).oucode_l2_code
+      find_court_centre_by_id(hearing_resulted.hearing_court_centre_id).oucode_l2_code
     end
 
     def cjs_location
@@ -26,7 +23,7 @@ module MaatApi
     end
 
     def case_creation_date
-      @shared_time.to_date.strftime("%Y-%m-%d")
+      hearing_resulted.shared_time.to_date.strftime("%Y-%m-%d")
     end
 
     def doc_language
@@ -45,6 +42,10 @@ module MaatApi
 
     def function_type
       "APPLICATION"
+    end
+
+    def jurisdiction_type
+      hearing_resulted.hearing_jurisdiction_type
     end
 
     def defendant
@@ -108,11 +109,11 @@ module MaatApi
     end
 
     def court_centre_short_ou_code
-      find_court_centre_by_id(hearing.court_centre_id).short_oucode
+      find_court_centre_by_id(hearing_resulted.hearing_court_centre_id).short_oucode
     end
 
     def hearing_first_sitting_day_date
-      hearing.first_sitting_day_date&.to_date&.strftime("%Y-%m-%d")
+      hearing_resulted.hearing_first_sitting_day_date&.to_date&.strftime("%Y-%m-%d")
     end
 
     def find_court_centre_by_id(id)
--- a/app/services/hearings_creator.rb
+++ b/app/services/hearings_creator.rb
@@ -15,20 +15,25 @@ class HearingsCreator < ApplicationService
 private
 
   def push_prosecution_cases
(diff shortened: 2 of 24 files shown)


spec/models/maat_api/court_application_spec.rb
```diff
@@ -1,111 +1,123 @@
 RSpec.describe MaatApi::CourtApplication, type: :model do
-  let(:hearing_body) { JSON.parse(file_fixture("hearing/with_court_application.json").read).deep_symbolize_keys }
+  let(:shared_time) { "2021-03-09T15:54:09.871Z" }
```
arthurashman, line 2: I don't think this is being used in this file

spec/models/maat_api/prosecution_case_spec.rb
```diff
@@ -1,12 +1,21 @@
 RSpec.describe MaatApi::ProsecutionCase, type: :model do
+  let(:shared_time) { "2021-03-09T15:54:09.871Z" }
```
arthurashman, line 2: I don't think this is being used in this file

app/services/hearings_creator.rb
```diff
@@ -15,48 +15,49 @@ def call
 private
 
   def push_prosecution_cases
-    hearing[:prosecutionCases]&.each do |prosecution_case_data|
-      prosecution_case_data[:defendants].each do |defendant_data|
-        next if defendant_data[:offences].any? { |offence| offence.dig(:laaApplnReference, :applicationReference)&.start_with?("A", "Z") }
+    hearing_resulted.hearing.prosecution_cases.each do |prosecution_case|
+      prosecution_case.defendants.each do |defendant|
+        next if defendant.offences.any? do |offence|
+          LaaReference.new(maat_reference: offence.laa_reference_application_reference).dummy_maat_reference?
```
arthurashman, line 21: This feels a bit OTT to me considering the info we need is on one of the parameters we're passing in to instantiate the LaaReference

app/services/hearings_creator.rb
```diff
@@ -15,48 +15,49 @@ def call
...
+        end
 
-        laa_reference = LaaReference.find_by!(defendant_id: defendant_data[:id], linked: true)
+        laa_reference = LaaReference.find_by(defendant_id: defendant.id, linked: true)
 
-        prosecution_case = MaatApi::ProsecutionCase.new(
-          hearing_body,
-          prosecution_case_data[:prosecutionCaseIdentifier][:caseURN],
-          defendant_data,
+        next if laa_reference.blank?
```
arthurashman, line 26: Is this actually correct? I know we discussed it. But this refactor does actually change what we send to the queue.

app/models/maat_api/court_application.rb
```diff
@@ -107,11 +109,11 @@ def offence_wording
     end
 
     def court_centre_short_ou_code
-      find_court_centre_by_id(hearing.court_centre_id).short_oucode
+      find_court_centre_by_id(hearing_resulted.hearing_court_centre_id).short_oucode
     end
 
     def hearing_first_sitting_day_date
-      hearing.first_sitting_day_date&.to_date&.strftime("%Y-%m-%d")
+      hearing_resulted.hearing.first_sitting_day_date&.to_date&.strftime("%Y-%m-%d")
```
arthurashman, line 116: Why not `hearing_resulted.hearing_first_sitting_day_date...` with delegate and prefex for `first_sitting_day_date`?

arthurashman: So it follows the same pattern as jurisdiction_type and court_centre_id

app/services/hearings_creator.rb
laurentqro: Yeah I know it's not great, but the knowledge of the concept of a dummy reference should live in one class only ideally, and it currently lives in `LaaReference`. Any suggestions?

app/services/hearings_creator.rb
laurentqro: This makes sure we don't end up with failed jobs, but ideally we would not even check CDA's internal state at all.

arthurashman: OK, right. I think I got confused thinking that those with blank `laa_references` would still get sent to MAATAPI. But they'll just hit a NoMethodError when we try to do `laa_reference.maat_reference` on line 32

app/services/hearings_creator.rb
arthurashman: Yeah I agree. The only thought I have is to have a MaatReference class - but it would basically result in exactly the same line of code. 

Maybe one for a future refactor.

arthurashman (dismissed):

app/services/hearings_creator.rb
laurentqro: Not exactly. `next` will skip the rest of the code block and move onto the next item to iterate on (in this case, the next defendant in the list we're iterating over).

https://www.rubyguides.com/2019/09/ruby-next-break-keywords/

arthurashman: Right, yes

spec/models/maat_api/court_application_spec.rb
laurentqro: Good spot, thanks!

app/models/maat_api/court_application.rb
laurentqro: Yeah, good idea 👌🏻

arthurashman (approved):

chris-groves (approved):

PR #564, opened 2021-06-11
Set up static analysis

laurentqro:
Create a static analysis Github Action workflow.

### DIFF
--- a/.github/workflows/brakeman.yml
+++ b/.github/workflows/brakeman.yml
@@ -1,0 +1,13 @@
+name: Brakeman
+
+on: [push]
+
+jobs:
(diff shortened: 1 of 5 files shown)


chris-groves (approved):

.github/workflows/brakeman.yml
```diff
@@ -0,0 +1,13 @@
+name: Brakeman
```
arthurashman, line 1: is this just the default config from running `brakeman -C` ?

laurentqro: I think this is just the name of the GH workflow

PR #582, opened 2021-06-28
LASB-623 update V2 GET Prosecution Case Swagger doc

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=454&projectKey=LASB&modal=detail&selectedIssue=LASB-623

Update the API documentation so users know what they will receive when querying CDA endpoint

## Checklist

### DIFF
--- a/spec/requests/api/internal/v2/prosecution_cases_spec.rb
+++ b/spec/requests/api/internal/v2/prosecution_cases_spec.rb
@@ -41,8 +41,9 @@ RSpec.describe "api/internal/v2/prosecution_cases", type: :request, swagger_doc:
 
           parameter name: "include", in: :query, required: false, type: :string,
                     schema: {},
-                    description: 'Can return defendant and hearing summary data as they both have a relationship with prosecution case. Offences, defence organisation and prosecution case data is related to defendants so in turn can also be included. </br>
-                                  eg include=defendants,hearing_summaries,defendants.offences,'
+                    description: 'Include top-level and nested associations for a prosecution case.
+                                  All top-level and nested associations available for inclusion are listed under the relationships keys of the response body.
+                                  e.g. to include hearing summaries as well as defendants and their offences: include=hearing_summaries,defendants,defendants.offences </br>
 
           parameter "$ref" => "#/components/parameters/transaction_id_header"
 
@@ -50,7 +51,7 @@ RSpec.describe "api/internal/v2/prosecution_cases", type: :request, swagger_doc:
 
           let(:Authorization) { "Bearer #{token.token}" }
           let(:'filter[prosecution_case_reference]') { "19GD1001816" }
-          let(:include) { "defendants,hearing_summaries,defendants.offences" }
+          let(:include) { "hearing_summaries,defendants,defendants.offences" }
 
           run_test!
         end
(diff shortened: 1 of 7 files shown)


spec/requests/api/internal/v2/prosecution_cases_spec.rb
```diff
@@ -41,16 +41,16 @@
 
           parameter name: "include", in: :query, required: false, type: :string,
                     schema: {},
-                    description: 'Return defendant, offence and hearing data through a has_many relationship </br>
-                                  eg include=defendants,defendants.offences,defendants.defence_organisation,hearing_summaries'
+                    description: 'Can return defendant and hearing summary data as they both have a relationship with prosecution case. Offences, defence organisation and prosecution case data is related to defendants so in turn can also be included. </br>
+                                  eg include=defendants,hearing_summaries,defendants.offences,'
```
laurentqro, line 45: Suggested wording: `Include top-level and nested associations for a prosecution case. All top-level and nested associations available for inclusion are listed under the relationships key of the response body. e.g. to include defendants and their offences: include=defendants,defendants.offences.`

Thoughts?

laurentqro:
Looking good @chris-groves. Could you look adding examples where they're missing? There are quite a few.

e.g. filter[arrest_summons_number] for the prosecution cases endpoint.

spec/requests/api/internal/v2/prosecution_cases_spec.rb
chris-groves: I like it!

swagger/v2/defendant.json
```diff
@@ -60,33 +60,48 @@
...
-      "description": "The LAA issued reference to the application. CDA expects a numeric number, although HMCTS allows strings",
-      "anyOf":[
-        {
-          "$ref": "laa_reference.json#/definitions/maat_reference"
-        },{
-          "type": "null"
-        }
-      ]
-    },
+     "example": 314159265,
```
arthurashman, line 63: Has rubocop changed indentation to 1 space? This is no longer consistent with the rest of the file's indentation

chris-groves: good spot!

swagger/v2/prosecution_case.json
```diff
@@ -110,6 +143,11 @@
         "data"
       ]
     },
+    "example_included_query_parameters": {
+      "description": "an example of parameters that can be included in the query",
+      "type": "string",
+      "example": "hearing_summaries,defendants,defendants.offences"
```
arthurashman, line 149: What are your thoughts on just having one example? Or if you want to indicate that a user can have multiple includes and needs a comma with no space - can we have two short examples? At the moment it runs over the example box length, and I don't think it can be expanded on swaggerUI

arthurashman: I don't personally feel like this example necessarily needs to match the example outlined in the description

spec/requests/api/internal/v2/prosecution_cases_spec.rb
```diff
@@ -40,17 +40,20 @@
                     description: "Searches prosecution cases by prosecution case reference"
 
           parameter name: "include", in: :query, required: false, type: :string,
-                    schema: {},
-                    description: 'Return defendant, offence and hearing data through a has_many relationship </br>
-                                  eg include=defendants,defendants.offences,defendants.defence_organisation,hearing_summaries'
+                    schema: {
```
arthurashman, line 43: Elsewhere in this file, there is inconsistency in naming the search variable names as part of the description. I wonder if this PR would be a good time to tidy them. 

`description: "Searches prosecution cases by national_insurance_number"`
vs
`description: "Searches prosecution cases by arrest summons number"`

My preference would be to not have the underscores, but I don't mind as long as they're all consistent

arthurashman (approved):

PR #589, opened 2021-07-02
Revert "Remove dependency on Court Centre data lookup gem"

laurentqro:
Reverts ministryofjustice/laa-court-data-adaptor#584


msahilan07 (approved):
LGTM

PR #593, opened 2021-07-08
Update V2 GET Defendants endpoint swagger doc

chris-groves:
## What

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=454&projectKey=LASB&modal=detail&selectedIssue=LASB-623

Update the API documentation for V2 GET defendants endpoint so users know what they will receive when querying this endpoint

## Checklist

### DIFF
--- a/swagger/v2/defence_organisation.json
+++ b/swagger/v2/defence_organisation.json
@@ -15,9 +15,9 @@
     "type": {
       "description": "The defence_organisations type",
       "enum": [
-        "defence_organisations"
+        "defence_organisation"
       ],
-      "example": "defence_organisations",
+      "example": "defence_organisation",
       "type": "string"
     },
     "name": {
@@ -27,26 +27,32 @@
     },
     "address1": {
       "readOnly": true,
+      "example": "104",
       "type": "string"
     },
     "address2": {
       "readOnly": true,
+      "example": "Road Street",
       "type": "string"
     },
     "address3": {
       "readOnly": true,
+      "example": "Westminster",
       "type": "string"
     },
     "address4": {
       "readOnly": true,
+      "example": "London",
       "type": "string"
     },
     "address5": {
       "readOnly": true,
+      "example": "UK",
       "type": "string"
     },
     "postcode": {
       "readOnly": true,
+      "example": "SW1 1AB",
       "type": "string"
     },
     "identity": {
@@ -56,11 +62,11 @@
       "description": "object representing a single defence_organisation",
       "type": "object",
       "properties": {
-        "type": {
-          "$ref": "#/definitions/type"
-        },
         "id": {
           "$ref": "#/definitions/id"
+        },
+        "type": {
+          "$ref": "#/definitions/type"
         },
         "attributes": {
           "$ref": "#/definitions/attributes"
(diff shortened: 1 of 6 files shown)


swagger/v2/defence_organisation.json
```diff
@@ -15,9 +15,9 @@
     "type": {
       "description": "The defence_organisations type",
       "enum": [
-        "defence_organisations"
+        "defence_organisation"
```
arthurashman, line 18: Is this not plural as in the comment from hearings that you discussed with me:

```
"type": {
      "description": "The hearing type",
      "example": "hearings",
      "type": "string"
    },
```

swagger/v2/defence_organisation.json
```diff
@@ -15,9 +15,9 @@
...
       ],
-      "example": "defence_organisations",
+      "example": "defence_organisation",
```
arthurashman, line 20: And here?

swagger/v2/prosecution_case.json
```diff
@@ -205,11 +215,11 @@
     "prosecution_case_reference": {
       "readOnly": true,
       "example": "TFL12345",
-      "description": "The prosecuting authorities reference for their prosecution case that is layed before court.  For example PTI-URN from police/cps cases",
+      "description": "The prosecuting authorities' reference for their prosecution case that is layed before court.  For example PTI-URN from police/cps cases",
```
arthurashman, line 218: I think this should be `authority's` - Also, could delete the second space between sentences, and I think laid is the more normal spelling not layed

swagger/v2/prosecution_case.json
```diff
@@ -157,20 +166,21 @@
...
       "description": "The prosecution cases type",
-      "enum": [
-        "prosecution_cases"
-      ],
-      "example": "prosecution_cases",
+      "example": "prosecution_case",
       "type": "string"
     },
+    "included_type": {
+      "description": "Prosecution case is referred to in the plural form when displayed in the payload's included section ",
```
arthurashman, line 181: Up for deleting the space at the end of the sentence?

arthurashman (commented):
Mostly looks good. tiny questions. 

Thank you for rewording the descriptions to proper sentences!

swagger/v2/defence_organisation.json
chris-groves: Good spot - will correct

chris-groves: I've sorted this by referencing a new "included_resource" object within the same file.  In the payload, it is pluralised when in the "included" but not when listed as a relationship. Having a separate object accommodates both scenarios.

swagger/v2/defence_organisation.json
chris-groves: See above

swagger/v2/prosecution_case.json
chris-groves: Good spot - I've corrected that

swagger/v2/defence_organisation.json
arthurashman: 👍

arthurashman: I wouldn't worry too much about defence organisation. We need to stop offering this because it is coming from CDA data, not Common Platform. So it's likely we'll be removing the option from the api - needs discussing further

arthurashman (approved):

hkhela (approved):

PR #599, opened 2021-07-09
Remove old HMCTS TLS client certificate from ingress allow list

laurentqro:
⚠️  Merge only when it is confirmed that HMCTS Common Platform has switched over to the new TLS client certificate.


arthurashman (approved):

laurentqro:
HMCTS have now told us that they are using the new certificate. This can be merged.

PR #632, opened 2021-08-12
Move service files into CP API folder

chris-groves:
Move service files with classes that interact with Common Platform into common_platform_api folder

## Checklist


app/services/common_platform_api/common_platform_connection.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module CommonPlatformApi
+  class CommonPlatformConnection < ApplicationService
```
arthurashman, line 4: What do you think about renaming this class if the module names it effectively now?

laurentqro: How about an `Api` and a `Connection` submodule under a `CommonPlatform` module?

arthurashman (approved):

PR #662, opened 2021-09-23
Expose judicial results

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-668)

### User Story 1

As A Caseworker

I want sentencing details on View Court Data

So That I can assess a bill without using multiple sources of information

### User Story 2 

As An Assess a claim developer 

I want to know where and how sentence related results are exposed to View Court Data

So That I can do my job and display this on the front end

### Acceptance Criteria

Scenario 1: CDA exposes necessary information to VCD

Given the results have been located on the payload 

When they are exposed to VCD

Then a VCD developer knows how to locate the relevant attributes in the payload

### DIFF
--- a/spec/serializers/api/internal/v1/offence_serializer_spec.rb
+++ b/spec/serializers/api/internal/v1/offence_serializer_spec.rb
@@ -1,7 +1,8 @@
 RSpec.describe Api::Internal::V1::OffenceSerializer do
   let(:serialized_data) do
+    offence_summary_data = JSON.parse(file_fixture("offence_summary/all_fields.json").read)
     offence_data = JSON.parse(file_fixture("offence/all_fields.json").read)
-    offence = Offence.new(body: offence_data, details: [offence_data])
+    offence = Offence.new(body: offence_summary_data, details: [offence_data])
 
     described_class.new(offence).serializable_hash[:data]
   end
@@ -11,7 +12,7 @@ RSpec.describe Api::Internal::V1::OffenceSerializer do
       let(:attributes) { serialized_data[:attributes] }
 
       it "code" do
-        expect(attributes[:code]).to eql("LA12505")
+        expect(attributes[:code]).to eql("TH68026C")
       end
 
       it "order_index" do
@@ -19,15 +20,15 @@ RSpec.describe Api::Internal::V1::OffenceSerializer do
       end
 
       it "title" do
-        expect(attributes[:title]).to eql("Driver / other person fail to immediately move a vehicle from a cordoned area on order of a constable")
+        expect(attributes[:title]).to eql("Conspire to commit a burglary dwelling with intent to steal")
       end
 
       it "legislation" do
-        expect(attributes[:legislation]).to eql("Common law")
+        expect(attributes[:legislation]).to eql("Contrary to section 1(1) of the    Criminal Law Act 1977.")
       end
 
       it "mode_of_trial" do
-        expect(attributes[:mode_of_trial]).to eq("Either way")
+        expect(attributes[:mode_of_trial]).to eq("Indictable")
       end
 
       it "mode_of_trial_reasons" do


spec/serializers/api/internal/v1/defendant_serializer_spec.rb
```diff
@@ -1,41 +1,81 @@
...
-                    representation_order_date: "2020-12-12",
-                    offence_ids: %w[55555],
-                    defence_organisation_id: "88888",
-                    prosecution_case_id: "5edd67eb-9d8c-44f2-a57e-c8d026defaa4",
-                    post_hearing_custody_statuses: %w[A])
-  end
+    let(:defendant) do
+      defendant_summary_data = JSON.parse(file_fixture("defendant_summary/all_fields.json").read)
+      defendant_data = JSON.parse(file_fixture("defendant/all_fields.json").read)
+      Defendant.new(body: defendant_summary_data, details: [defendant_data], prosecution_case_id: "123")
```
arthurashman: These 3 lines make all of this much clearer, thank you!

spec/serializers/api/internal/v1/offence_serializer_spec.rb
```diff
@@ -1,29 +1,42 @@
...
-                    code: "AA06001",
-                    order_index: "0",
-                    title: "Fail to wear protective clothing",
-                    legislation: "Offences against the Person Act 1861 s.24",
-                    mode_of_trial: "Indictable-Only Offence",
-                    mode_of_trial_reasons: [{ description: "Court directs trial by jury", code: "5" }],
-                    pleas: %w[GUILTY NOT_GUILTY])
+  let(:serialized_data) do
+    offence_data = JSON.parse(file_fixture("offence/all_fields.json").read)
+    offence = Offence.new(body: offence_data, details: [offence_data])
```
arthurashman, line 4: Does the body here not need to come from offence_summary_data?

app/models/defendant.rb
```diff
@@ -55,6 +55,16 @@ def defence_organisation_id
     defence_organisation&.id
   end
 
+  def judicial_result_ids
+    judicial_results.map(&:id)
+  end
+
+  def judicial_results
```
arthurashman, line 62: We were considering getting defendant JRs from Hearing>defendantJudicialResults
Are you happy that we're surfacing them here instead?

app/models/defendant.rb
```diff
@@ -77,6 +87,12 @@ def offence_details
     details.flat_map { |detail| detail["offences"] }.group_by { |offence| offence["id"] }
   end
 
+  def judicial_results_array
+    return [] if details.blank?
+
+    details.flat_map { |detail| detail["judicialResults"] }.uniq.compact
```
arthurashman, line 93: Do we know that we want these to be unique this time? I don't know the domain well enough to know why pleas was an array of unique elements...

laurentqro: I think that's just in case judicial results are repeated for a given defendant. This makes sure we end up with a list of uniques.

app/models/defendant.rb
laurentqro: Might be worth double-checking, but I think both ways would yield the same result.

spec/serializers/api/internal/v1/defendant_serializer_spec.rb
laurentqro: my pleasure!

spec/serializers/api/internal/v1/offence_serializer_spec.rb
laurentqro: yes, good catch! thank you

app/models/defendant.rb
arthurashman: I couldn't see a hearing that had a defendantJudicial results when I checked post writing this comment, so maybe this way is best on reflection!

arthurashman (approved):

hkhela (approved):

arthurashman (approved):

chris-groves (approved):

PR #683, opened 2021-11-01
Add defendantId to unlinking payload

laurentqro:
## What

Add defendant ID to the message that gets sent to MAAT API when a case worker unlinks a defendant.

## Why

This is in preparation of removing CDA's dependency on its `laa_references` table, used to send `maatId` to MAAT. Going forward, MAAT API will be sole responsible for looking up maat id for a defendant id.

### DIFF
--- a/app/services/laa_reference_unlinker.rb
+++ b/app/services/laa_reference_unlinker.rb
@@ -24,6 +24,7 @@ private
   def push_to_queue
     Sqs::MessagePublisher.call(
       message: {
+        defendantId: defendant_id,
         maatId: laa_reference.maat_reference,
         userId: user_name,
         reasonId: unlink_reason_code,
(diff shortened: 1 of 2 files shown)


app/services/laa_reference_unlinker.rb
```diff
@@ -24,6 +24,7 @@ def unlink_maat_reference!
   def push_to_queue
     Sqs::MessagePublisher.call(
       message: {
+        defendantId: defendant_id,
         maatId: laa_reference.maat_reference,
```
arthurashman, line 28: So we're still sending MAAT ID?

laurentqro: two-prong approach:

1. add defendantId and if everything is fine
2. remove maatId

atharmajeed (approved):
looks good to me!

PR #684, opened 2021-11-01
Expose hearingSummary.hearingDay.hasSharedResults flag

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-718)

CDA will need to expose the status of results shared for VCD to consume during the search operation. This information will be available within the hearingSummary.hearingDay.hasResultsShared  as part of HMCTS’ search API response.  More detail can be found on this link

## Why

This information will be displayed on the VCD screen which will enable billing caseworkers to understand the status of the results shared. 

## Definition of Done

* [x] Results shared status is made available within CDA’s search API & tested.
* [x] The relevant API docs are updated to reflect this new info. 
* [x] v1/v2 services are updated.


chris-groves (commented):
Looks good. 

Is it worth adding tests for the hearing_summary model and serializer?

The has_shared_results field renders between the hearing_type and hearing_days fields, whereas the swagger documentation states that has_shared_results will come after hearing_type and hearing_days.

arthurashman:
The docs say that hasSharedResults is on the HeargingSummary>HearingDay, not the HearingSummary

app/models/hearing_summary.rb
```diff
@@ -39,8 +39,6 @@ def court_centre
     HmctsCommonPlatform::CourtCentre.new(body["courtCentre"])
   end
 
-private
-
   def hearing_days
```
arthurashman, line 42: Can we have a test for this?

app/models/hearing_summary.rb
```diff
@@ -39,8 +39,6 @@ def court_centre
     HmctsCommonPlatform::CourtCentre.new(body["courtCentre"])
   end
```
arthurashman, line 41: models/hearing_summary.rb has a resulted? method. Wondering if the logic for calculating this is faulty/outdated, and we could now use the hasSharedResults attribute?

arthurashman (approved):
Looks good. Thank you for doing the refactor work!

PR #690, opened 2021-11-08
Expose DefenceCounsel on prosecution case search response

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-720)

Exposes Defence Counsel information for VCD to consume during the prosecution search operation.

This information will be available within the `hearingSummary.DefenseCounsel` as part of HMCTS’ prosecution case search API response.

### DIFF
--- a/app/models/hearing_summary.rb
+++ b/app/models/hearing_summary.rb
@@ -43,11 +43,21 @@ class HearingSummary
     HmctsCommonPlatform::CourtCentre.new(body["courtCentre"])
   end
 
+  def defence_counsel_ids
+    defence_counsels.map(&:id)
+  end
+
 private
 
   def hearing_days
     Array(body["hearingDays"]).map do |hearing_day_data|
       HmctsCommonPlatform::HearingDay.new(hearing_day_data)
+    end
+  end
+
+  def defence_counsels
+    Array(body["defenceCounsel"]).map do |defence_counsel_data|
+      HmctsCommonPlatform::DefenceCounsel.new(defence_counsel_data)
     end
   end
 end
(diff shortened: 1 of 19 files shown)


app/models/hearing_summary.rb
```diff
@@ -43,11 +43,21 @@ def court_centre
...
 private
 
   def hearing_days
     Array(body["hearingDays"]).map do |hearing_day_data|
       HmctsCommonPlatform::HearingDay.new(hearing_day_data)
     end
   end
+
+  def defence_counsels
+    Array(body["defenceCounsel"]).map do |defence_counsel_data|
```
arthurashman, line 59: Do you have a proper updated schema? I'm only going off the diagram in the solution paper, but it says there's 0 or 1 defence counsel on there

chris-groves (approved):

app/serializers/api/internal/v1/hearing_summary_serializer.rb
```diff
@@ -18,6 +19,8 @@ class HearingSummarySerializer
             name: hearing_summary.court_centre.name,
           }
         end
+
+        has_many :defence_counsels
       end
```
chris-groves, line 24: Can you update the request specs so the swagger UI mentions that we can include hearing_summaries.defence_counsel in the api call to the GET prosecution cases endpoint?

chris-groves (approved):

app/models/hearing_summary.rb
laurentqro: the `0..1` is near `HearingSummary`, and the `DefenceCousel` has a `*`. Not entirely sure what that means, but I interpret it as `0 or 1 hearing summary has many DefenceCounsel`.

arthurashman: Right I see that now. Their way of allowing the diagram to show if it's a one to many or many to many relationship I guess

PR #693, opened 2021-11-10
Upgrade ruby and postgres CircleCI images

chris-groves:
## What

[[Link to story](https://dsdmoj.atlassian.net/browse/-LASBXXX)](https://dsdmoj.atlassian.net/jira/software/c/projects/LASB/boards/454?modal=detail&selectedIssue=LASB-732)

Upgrade Ruby and Postgres CircleCI images so they continue to receive bug/security fixes; the legacy images will no longer be supported from the end of 2021.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
           PAGER: cat
           DATABASE_URL: postgres://postgres@localhost/laa_court_data_adaptor_test
           RAILS_ENV: test
-      - image: cimg/postgres:12.8
+      - image: cimg/postgres:11.12
         environment:
           POSTGRES_DB: laa_court_data_adaptor_test
     steps:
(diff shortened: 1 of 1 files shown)


laurentqro (dismissed):

Mhowell494:
Is the Postgres version used in production the same as the one we're testing against? I feel like that's something we should consider.

laurentqro:
@Mhowell494 good point. We use 11.12.

```
irb(main):002:0> ActiveRecord::Base.connection.select_value('SELECT version()')
=> "PostgreSQL 11.12 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.3 20140911 (Red Hat 4.8.3-9), 64-bit"
```

laurentqro (approved):

PR #710, opened 2021-11-30
Inject queue URL from the worker

laurentqro:
This makes it easy to switch queue URL for the HearingsCreator
at runtime.

e.g. my use case: Fetch hearing results from Common Platform and send
the hearing resulted data to the prosecution_concluded queue rather
than the hearing_resulted queue.


arthurashman (approved):

arthurashman:
All makes sense. LGTM

PR #740, opened 2022-01-13
Fix error no method "urn"

laurentqro:
## What

Fix https://sentry.io/organizations/ministryofjustice/issues/2913986575/events/3c6c22330e2b4592afa83ecf513f72fc/?environment=prod&project=5375870.


msahilan07 (approved):

arthurashman (approved):

arthurashman (commented):
Wait, actually, could this not break VCD if they're using urn?

PR #765, opened 2022-01-21
add sitting_day param to hearing request made to Common Platform

arthurashman:
* Add sitting_day param to hearing request made to Common Platform as part of CDA's Get Prosecution Case endpoint
* Add conditional logic to only make request for resulted hearings
* Update Swagger Docs
* Update both API v1 and v2

### DIFF
--- a/app/controllers/api/internal/v1/hearing_results_controller.rb
+++ b/app/controllers/api/internal/v1/hearing_results_controller.rb
@@ -5,20 +5,30 @@ module Api
     module V1
       class HearingResultsController < ApplicationController
         def show
-          @hearing = CommonPlatform::Api::GetHearingResults.call(hearing_id: params[:id])
+          @hearing = CommonPlatform::Api::GetHearingResults.call(
+            hearing_id: permitted_params[:id],
+            sitting_day: permitted_params[:sitting_day],
+          )
+
           render json: HearingSerializer.new(@hearing, serialization_options)
         end
 
       private
 
         def serialization_options
-          return { include: inclusions } if inclusions.present?
-
-          {}
+          { include: inclusions }.compact
         end
 
         def inclusions
-          params[:include]&.split(",")
+          permitted_params[:include]&.split(",")
+        end
+
+        def permitted_params
+          params.permit(
+            :id,
+            :sitting_day,
+            :include,
+          )
         end
       end
     end
--- a/app/services/common_platform/api/hearing_fetcher.rb
+++ b/app/services/common_platform/api/hearing_fetcher.rb
@@ -5,8 +5,8 @@ module CommonPlatform
     class HearingFetcher < ApplicationService
       URL = "hearing/results"
 
-      def initialize(hearing_id:, connection: CommonPlatform::Connection.call)
-        @params = { hearingId: hearing_id }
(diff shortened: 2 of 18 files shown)


pking-moj (dismissed):

mabely (dismissed):

pking-moj (dismissed):

ivanELEC (dismissed):

app/controllers/api/internal/v1/hearing_results_controller.rb
```diff
@@ -5,20 +5,30 @@ module Internal
     module V1
       class HearingResultsController < ApplicationController
         def show
-          @hearing = CommonPlatform::Api::GetHearingResults.call(hearing_id: params[:id])
+          @hearing = CommonPlatform::Api::GetHearingResults.call(
+            hearing_id: permitted_params[:id],
+            sitting_day: permitted_params[:sitting_day],
+          )
+
           render json: HearingSerializer.new(@hearing, serialization_options)
```
arthurashman, line 13: I know we're deprecating v1, but would it be worth wrapping the render with the conditional guard like we have in v2?

```
if hearing_result.present?
            render json: HmctsCommonPlatform::HearingResulted.new(hearing_result&.body).to_json,
                   status: :ok
          else
            render json: {}, status: :not_found
          end
```

app/services/common_platform/api/hearing_fetcher.rb
```diff
@@ -6,7 +6,7 @@ class HearingFetcher < ApplicationService
       URL = "hearing/results"
 
       def initialize(hearing_id:, sitting_day:, connection: CommonPlatform::Connection.call)
-        @params = { hearingId: hearing_id, sittingDay: sitting_day }.compact
+        @params = { hearingId: hearing_id, sittingDay: sitting_day }.reject { |_k, v| v.blank? }
```
arthurashman, line 9: For my learning, can you explain the difference here?

laurentqro: `compact` rejects key/value pairs with a `nil` value only.

Rejecting all key/value pairs with a blank value means we also filter out empty strings.

```
irb> hash = {foo: "bar", baz: ""}

irb> hash.compact
=> {:foo=>"bar", :baz=>""}

irb)> hash.reject { |_k, v| v.blank? }
=> {:foo=>"bar"}
```

app/controllers/api/internal/v1/hearing_results_controller.rb
laurentqro: VCD are currently not using this endpoint at all. They will only start using the hearing result endpoint when they move to v2.

app/services/common_platform/api/hearing_fetcher.rb
arthurashman: Awesome, thank you!

app/controllers/api/internal/v1/hearing_results_controller.rb
arthurashman: Gotcha. I thought that this was the hearing endpoint renamed. All good

laurentqro (dismissed):

laurentqro (approved):

PR #804, opened 2022-02-07
Shift DNS weighting to 100% live cluster (prod namespace)

laurentqro:
## What

Shift DNS weighting to 100% live cluster


atharmajeed (approved):
👍🏼

PR #845, opened 2022-03-03
Pass hash as Sidekiq job argument, with string keys, to avoid using symbols

laurentqro:
This is to avoid using symbols, that are not native JSON types while at the same time not converting our hashes entirely into JSON.

From the Sidekiq [Wiki](https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple):

> The arguments you pass to perform_async must be composed of simple JSON datatypes:
string, integer, float, boolean, null(nil), array and hash. This means you must not
use ruby symbols as arguments. The Sidekiq client API uses JSON.dump to send the data
to Redis. The Sidekiq server pulls that JSON data from Redis and uses JSON.load to
convert the data back into Ruby types to pass to your perform method. Don't pass symbols,
named parameters, keyword arguments or complex Ruby objects (like Date or Time!) as those
will not survive the dump/load round trip correctly. You can use methods like
Hash#stringify_keys to convert Symbol keys when passing a Hash to perform_async.

Fixes issues https://github.com/ministryofjustice/laa-court-data-adaptor/issues/844 and https://github.com/ministryofjustice/laa-court-data-adaptor/issues/843 that were introduced by PR https://github.com/ministryofjustice/laa-court-data-adaptor/pull/841.

### DIFF
--- a/app/services/hearing_recorder.rb
+++ b/app/services/hearing_recorder.rb
@@ -51,7 +51,7 @@ private
   def publish_hearing_to_queue
     HearingsCreatorWorker.perform_async(
       Current.request_id,
-      hearing_resulted_data.to_json,
+      hearing_resulted_data.deep_stringify_keys,
     )
   end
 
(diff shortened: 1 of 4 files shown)


arthurashman (approved):
LGTM

PR #851, opened 2022-03-10
Rename hearing_results param from id to hearing_id

laurentqro:
## What

Rename hearing_results param from id to hearing_id


mabely (approved):
Looks great - thanks!

PR #868, opened 2022-03-22
Don't log full message (sensitive info)

laurentqro:
## What

Don't log full message that gets posted to the queue to avoid posting too much PII to Kibana.


laurentqro:
@msahilan07 yes I'll do that in a future PR, I'd like to stop logging PII first :)

PR #931, opened 2022-07-06
Expose defendant full name

laurentqro:
Fix error: `undefined method "name" for HmctsCommonPlatform::Defendant`.

This PR fixes the error by renaming `name` to `full_name` on the DefendantSerializer and in all places necessary.

### DIFF
--- a/app/models/hmcts_common_platform/person.rb
+++ b/app/models/hmcts_common_platform/person.rb
@@ -39,6 +39,7 @@ module HmctsCommonPlatform
     def nino
       data[:nationalInsuranceNumber]
     end
+    alias_method :national_insurance_number, :nino
 
     def occupation
       data[:occupation]


app/models/hmcts_common_platform/defendant.rb
```diff
@@ -77,5 +104,16 @@ def to_builder
...
+    def valid_maat_reference?
+      unique_maat_reference.present? && !unique_maat_reference.start_with?("Z")
+    end
+
+    def unique_maat_reference
+      refs = offences.map(&:laa_application_reference).uniq.compact
+      raise "Too many maat references" if refs.size > 1
+
+      refs&.first
+    end
```
arthurashman, line 117: Could these be refactored? Maybe combined? It looks like `unique_maat_reference` will get called 3 times for each call of `maat_reference`

PR #1018, opened 2022-10-27
A duplicate linking request unlinks most recent link before creating a new one

laurentqro:
## What

## Checklist

### DIFF
--- a/app/services/maat_link_creator.rb
+++ b/app/services/maat_link_creator.rb
@@ -50,7 +50,15 @@ private
   end
 
   def persist_laa_reference
-    laa_reference.save
+    existing_laa_ref = LaaReference.find_by(
+      defendant_id: laa_reference.defendant_id,
+      maat_reference: laa_reference.maat_reference,
+      linked: true,
+    )
+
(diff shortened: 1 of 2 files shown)


app/services/maat_link_creator.rb
```diff
@@ -50,7 +50,15 @@ def post_laa_reference_to_common_platform(offence)
...
-    laa_reference.save
+    existing_laa_ref = LaaReference.find_by(
+      defendant_id: laa_reference.defendant_id,
+      maat_reference: laa_reference.maat_reference,
+      linked: true,
+    )
+
+    existing_laa_ref.unlink! if existing_laa_ref.present?
+
+    laa_reference.save!
```
Obsiye, line 61: Hi, LGTM. Just a question about what happens on MAAT side.

My understanding is when there's a duplicate link request, it still executes `#publish_laa_reference_to_queue`
, would that mean the MAAT app will have to check if a link exists, unlink and link the new one?

e.g.

**CDA**

1. link exists
2. "unlink" 
3. link again

**MAAT**

1. link exists , will it throw an error? ("Link already exists")

Obsiye (approved):

PR #1075, opened 2023-01-23
Create rake task to fetch past hearings for given MAAT IDs

laurentqro:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-1640)

Create a rake tasks that allows fetching past hearings for a list of MAAT IDs.

This is useful at times because LAA does not reliably receive all hearings from HMCTS, which means LAA is required to fetch past hearings once in a while to fill the gaps.


mtac50 (approved):
Looks good!

PR #1089, opened 2023-02-16
Improvements on the docker-compose and README

alexdesi:
https://dsdmoj.atlassian.net/browse/LASB-1714

## What
- Add info in README.md on how to run app with docker-compose
- Fix docker-compose database connection

## Checklist

### DIFF
--- a/.env
+++ b/.env
@@ -14,3 +14,4 @@ MAAT_API_OAUTH_URL=https://maat-cd-api-dev.auth.eu-west-2.amazoncognito.com
 MAAT_API_API_URL=https://maat-cd-api.dev.legalservices.gov.uk
 MAAT_API_CLIENT_ID=
 MAAT_API_CLIENT_SECRET=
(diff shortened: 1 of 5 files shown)


mtac50 (dismissed):

.env
```diff
@@ -14,3 +14,5 @@ MAAT_API_OAUTH_URL=https://maat-cd-api-dev.auth.eu-west-2.amazoncognito.com
 MAAT_API_API_URL=https://maat-cd-api.dev.legalservices.gov.uk
 MAAT_API_CLIENT_ID=
 MAAT_API_CLIENT_SECRET=
+DATABASE_URL=postgres://postgres:postgres@postgres:5432/laa_court_data_adaptor_development
```
mtac50, line 17: Technically this wouldn't this need to go in the `.env.development.local` env file? Having it here means the build for the production image may pick it up.

alexdesi: Good point, not sure if  the value of DATABASE_URL is replaced when deployed, but as you said it is enough to have it on the local env.

mtac50 (approved):

PR #1095, opened 2023-02-27
Lasb 1747 fix prometheus on dev

alexdesi:
[Link to story](https://dsdmoj.atlassian.net/browse/-LASB-1747)

## What
LASB-1747: disable prometheus export in non-production environments.
    This change has been made because we don't need to collect data on development/local env,
    and to suppress the errors in the docker-compose terminal.

## Checklist


alexdesi:
Screenshot 2023-02-27 at 13 43 14

mtac50 (commented):
Is there any way to limit this so the condition is something like `if not Rails.env.local`? I was just thinking that we do check the Prometheus data when we manually test in other environments

alexdesi:
> Is there any way to limit this so the condition is something like `if not Rails.env.local`? I was just thinking that we do check the Prometheus data when we manually test in other environments

I guess that is not a problem, you can still check prometheus on testing env:
I t might seem strange, but even in the test env,
`Rails.env.production?` is `true`

To check it, you can show the env var in the POD, for example:
`kubectl exec -it laa-court-data-adaptor-...  env -n laa-court-data-adaptor-test`

Also, I've changed it to  `if Rails.env.production?` 
to match the initializer: https://github.com/ministryofjustice/laa-court-data-adaptor/blob/90a00732ee2d3c949989ff30dc73561ddbf02168/config/initializers/prometheus.rb#L1

mtac50 (approved):

PR #1122, opened 2023-04-21
LASB-1908: This change stop raising exception every time the HearingF…

alexdesi:
## What

HearingFetcher returns success and empty body.

The reason is to avoid MaatLinkCreator to be run many times by the MaatLinkCreatorWorker, since sidekiq retries to run it when an exception occurs.

Story:
https://dsdmoj.atlassian.net/browse/LASB-1908

## Checklist


alexdesi:
This fix has been included in a different PR (LASB-1976-improve-hearings-event-log-endpoint).

PR #1125, opened 2023-05-02
LASB-1858: 

alexdesi:
## What
This test is very similar to the defendant_request_spec.
It is a controller test, and we want only test if the exception is Errors::DefendantError is rescued with 422 HTTP status.

## Checklist


spec/controllers/api/internal/v1/defendants_controller_spec.rb
```diff
@@ -8,39 +8,25 @@
...
-    # This case contain a `offenceSummary` having more than one applicationReference
-    let(:prosecution_case_reference) { "44PC1234567" }
-
     describe "GET #show" do
       it "raises HTTP 422 Unprocessable Entity" do
         get :show, params: { id: defendant_id }, as: :json
 
         expect(response).to have_http_status(:unprocessable_entity)
-        expect(response.body).to include("error\":\"Too many maat references")
+        expect(response.body).to include("error\":\"Too many maat references: [111,222]")
```
alexdesi, line 28: This is only testing that the exception is rescued, which is okay, because it is a controller.
The logic is tested here: https://github.com/ministryofjustice/laa-court-data-adaptor/blob/48ddd38fc2990ea578a077b0fa384751a9aa1576/spec/models/defendant_spec.rb#L144

PR #1127, opened 2023-05-05
Lasb 1925 fix maat link creation

alexdesi:
## What

This PR:
-  prevents Sidekiq to retries N times the MaatLinkCreator.call.
    It does it because:
    - it ensures that the laa ref is set to unlink, regardless of the defendant_id
    - it captures possible ActiveRecords exceptions and reports them to Sentry

-  prevents retries when CP replies 200 with an empty body.

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-1925)

Important: before deploying this, we must remove old jobs from Sidekiq. Ask @alexdesi for the scripts to run.

The changes in this PR have been discussed with @mtac50 

## Checklist

### DIFF
--- a/app/models/laa_reference.rb
+++ b/app/models/laa_reference.rb
@@ -11,18 +11,14 @@ class LaaReference < ApplicationRecord
 
   def adjust_link_and_save!
     # Unlink the previous laa reference
-    laa_ref_already_liked = LaaReference.find_by(maat_reference: maat_reference, linked: true)
-    laa_ref_already_liked.unlink! if laa_ref_already_liked.present?
+    laa_ref_already_linked = LaaReference.find_by(maat_reference: maat_reference, linked: true)
+    laa_ref_already_linked.unlink! if laa_ref_already_linked.present?
 
     self.linked = true
 
     save!
   rescue ActiveRecord::ActiveRecordError => e
     Sentry.capture_exception(e)
-  end
-
-  def self.unlink_all!(maat_ref)
-    where(maat_reference: maat_ref).update_all(linked: false)
   end
 
   def dummy_maat_reference?
--- a/app/models/laa_reference.rb
+++ b/app/models/laa_reference.rb
@@ -9,6 +9,22 @@ class LaaReference < ApplicationRecord
     update!(linked: false)
   end
 
+  def adjust_link_and_save!
+    # Unlink the previous laa reference
+    laa_ref_already_liked = LaaReference.find_by(maat_reference: maat_reference, linked: true)
+    laa_ref_already_liked.unlink! if laa_ref_already_liked.present?
+
+    self.linked = true
+
+    save!
+  rescue ActiveRecord::ActiveRecordError => e
+    Sentry.capture_exception(e)
+  end
+
+  def self.unlink_all!(maat_ref)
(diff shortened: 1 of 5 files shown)


app/models/laa_reference.rb
```diff
@@ -9,6 +9,18 @@ def unlink!
...
+    # Unlink the previous laa reference
+    laa_ref_already_liked = LaaReference.find_by(maat_reference: maat_reference, linked: true)
+    laa_ref_already_liked.unlink! if laa_ref_already_liked.present?
+
+    self.linked = true
+
+    save!
+  rescue ActiveRecord::ActiveRecordError => e
+    Sentry.capture_exception(e)
+  end
```
alexdesi, line 22: This is the method that fixes the issue:
The key concept is that the `find_by` no longer looks for a certain defendant_id. Any linked maat_ref must be unlinked, and then the current maat_ref can be linked.
In this way, we prevent the uniqueness validation (https://github.com/ministryofjustice/laa-court-data-adaptor/blob/d6e347f8d10f284bb943bc66fa6743a006de8994/app/models/laa_reference.rb#L5)

app/services/hearing_result_fetcher.rb
```diff
@@ -14,23 +14,33 @@ def call
...
     end
 
-    unless response.success?
-      msg += "Unable to fetch past result of hearing ID #{hearing_id}: Common Platform responded with status code #{response.status}."
-      Rails.logger.info(msg)
-      raise StandardError, msg
+    if (500..599).cover?(response.status) # Server Error 5XX
+      msg = log_for("Unable to fetch past result of hearing ID #{hearing_id}: Common Platform responded with status code #{response.status}.")
+
+      raise Faraday::ServerError, msg # this tells Sidekiq to Retry
```
alexdesi, line 37: it retries the hearing fetching only in case of server error, which is most of the time 503

app/services/maat_link_creator.rb
```diff
@@ -65,18 +66,6 @@ def post_laa_reference_to_common_platform(offence)
     raise StandardError, "Error posting LAA Reference to Common Platform" unless response.success?
   end
 
-  def persist_laa_reference
-    existing_laa_ref = LaaReference.find_by(
-      defendant_id: laa_reference.defendant_id,
-      maat_reference: laa_reference.maat_reference,
-      linked: true,
-    )
```
alexdesi, line 73: This method is the main culprit for the bug.
I've refactored it. It's now called `adjust_link_and_save!`

app/models/laa_reference.rb
mtac50: I was more thinking of the unlink process in the scenario where we are unlinking the maat_id from a different defendant?

app/models/laa_reference.rb
```diff
@@ -9,6 +9,18 @@ def unlink!
     update!(linked: false)
   end
 
+  def adjust_link_and_save!
+    # Unlink the previous laa reference
+    laa_ref_already_liked = LaaReference.find_by(maat_reference: maat_reference, linked: true)
+    laa_ref_already_liked.unlink! if laa_ref_already_liked.present?
```
alexdesi, line 15: fix this: laa_ref_already_li**N**ked

mtac50 (approved):

PR #1128, opened 2023-05-12
LASB-1976: Improve the error message in case of status error on CP API.

alexdesi:
## What
Improve the error message in case of a status error on Cloud Platform API.
Also, fix a rspec hearing_event_logs_request_spec.rb with proper. and rspec

## Checklist


spec/requests/api/internal/v2/hearing_event_logs_request_spec.rb
```diff
@@ -63,22 +63,28 @@
...
 
-      context "when no hearing log is found" do
+      context "when no Hearing ID found" do
         let(:hearing_id) { "c6cf04b5" }
         let(:date) { "2020-02-09" }
 
         before do
           stub_request(:get, "#{ENV['COMMON_PLATFORM_URL']}/hearing/hearingLog?hearingId=#{hearing_id}&date=#{date}")
             .to_return(
-              status: 200,
```
alexdesi, line 73: CP Api returns 500 when the Hearing is not found (not 200).
Note: this test used to work because it also returns an empty body.

app/controllers/api/internal/v2/hearing_event_logs_controller.rb
```diff
@@ -6,16 +6,31 @@ module V2
...
+          params[:hearing_id]
+        end
+
+        def hearing_date
+          params[:hearing_date]
+        end
+
+        def unsuccess_error_message
+          "The hearing '#{hearing_id}' on the '#{hearing_date}' was not found, " \
+            "or a Server Error occurred on the Cloud Platform API."
```
alexdesi, line 32: I added an error message to give more info to the caller.

mtac50, line 32:
```suggestion
            "or a Server Error occurred on the Common Platform API."
```

spec/requests/api/internal/v2/hearing_event_logs_request_spec.rb
```diff
@@ -63,22 +63,28 @@
...
 
         describe "responds with" do
           response(404, "not found") do
-            run_test!
+            run_test! do |response|
+              data = JSON.parse(response.body)
+
+              expect(data["error"])
+                .to eq("The hearing '#{hearing_id}' on the '#{date}' was not found, " \
+                       "or a Server Error occurred on the Cloud Platform API.")
```
mtac50, line 86:
```suggestion
                       "or a Server Error occurred on the Common Platform API.")
```

spec/requests/api/internal/v2/hearing_event_logs_request_spec.rb
```diff
@@ -63,22 +63,28 @@
         end
       end
 
-      context "when no hearing log is found" do
+      context "when no Hearing ID found" do
```
mtac50, line 66:
```suggestion
      context "when no data is found for the given Hearing ID" do
```

app/controllers/api/internal/v2/hearing_event_logs_controller.rb
alexdesi: Ops, good catch! thanks 😄

spec/requests/api/internal/v2/hearing_event_logs_request_spec.rb
alexdesi: This change is intentional: when the ID of the hearing is not found, it returns 500 errors.
In case no data is found for the hearing, it returns 200. 
We already have another test with 200: https://github.com/ministryofjustice/laa-court-data-adaptor/blob/2399a564f4a6d5ad5c6aa893c937603cc2878f53/spec/requests/api/internal/v2/hearing_event_logs_request_spec.rb#L35

Maybe I could add a new one with 200 and empty returns, but it does not add anything to the test on #L35

mtac50: Ah ok, thanks!

mtac50 (approved):

PR #1137, opened 2023-05-25
Filter the messages that are sent to the AWS_HEARING_RESULTED_QUEUE_URL

alexdesi:
## What
Filter the messages that are sent to the AWS_HEARING_RESUTTED_QUEUE_URL queue defendant_id.

I've added a new class to filter out the defendants that are not relevant.

## Why

Context is provided in the ticket: [Link to story](https://dsdmoj.atlassian.net/browse/LASB-1946)

## Checklist

- [ x] You should have checked that the commit messages say why the change was made.

### DIFF
--- a/app/services/common_platform/hearing_results_filter.rb
+++ b/app/services/common_platform/hearing_results_filter.rb
@@ -1,0 +1,20 @@
+module CommonPlatform
+  class HearingResultsFilter < ApplicationService
+    def initialize(hearing_results_body, defendant_id: nil)
+      @body = hearing_results_body.deep_stringify_keys
+      @defendant_id = defendant_id
+    end
+
+    def call
+      return @body unless @defendant_id
+
+      @body["hearing"]["prosecutionCases"] = @body["hearing"]["prosecutionCases"].map do |prosecution_case|
+        prosecution_case.merge(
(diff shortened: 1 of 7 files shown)


app/services/common_platform/hearing_results_filter.rb
```diff
@@ -0,0 +1,20 @@
...
+  class HearingResultsFilter < ApplicationService
+    def initialize(hearing_results_body, defendant_id: nil)
+      @body = hearing_results_body.deep_stringify_keys
+      @defendant_id = defendant_id
+    end
+
+    def call
+      return @body unless @defendant_id
+
+      @body["hearing"]["prosecutionCases"] = @body["hearing"]["prosecutionCases"].map do |prosecution_case|
```
mtac50, line 11: I think we'll also need to extend the filtering to the `court_applications`?

alexdesi: so maybe not. In that case,
`LaaReference.find_by(defendant_id: defendant.id, linked: true)` should filter for the only defendant that is linked.
see:
https://github.com/ministryofjustice/laa-court-data-adaptor/blob/acc9e845a59ddae97dac6a9bbeaebd0e005c146f/app/services/hearings_creator.rb#L27

nittagr (approved):

nittagr (approved):

PR #1148, opened 2023-06-08
Lasb 2034 add missing rspec to test SQS messages

alexdesi:
## What

Add missing rspec to test that **HearingsCreator** and **MaatLinkCreator** publish the proper messages to SQS.
[Link to story](https://dsdmoj.atlassian.net/browse/Lasb2034)

### DIFF
--- a/app/controllers/api/external/v1/prosecution_conclusions_controller.rb
+++ b/app/controllers/api/external/v1/prosecution_conclusions_controller.rb
@@ -8,12 +8,12 @@ module Api
           prosecution_conclusion_params["prosecutionConcluded"].each do |pc|
             laa_reference = LaaReference.find_by(defendant_id: pc["defendantId"], linked: true)
 
-            if laa_reference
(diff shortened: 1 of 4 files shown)


app/controllers/api/external/v1/prosecution_conclusions_controller.rb
```diff
@@ -5,15 +5,15 @@ class ProsecutionConclusionsController < ApplicationController
         def create
           enforce_contract!
 
-          prosecution_conclusion_params["prosecutionConcluded"].map do |pc|
+          prosecution_conclusion_params["prosecutionConcluded"].each do |pc|
```
alexdesi, line 8: I think this was an error, it does not make sense to have a `map` here.
I think who wrote this wanted to use an `each` loop. That's why there is a `next` inside (which I removed anyway).
By chance, it works because `map` run the code to create the array.
Fixed now.

mtac50 (approved):

PR #1149, opened 2023-06-09
LABS-2037: Improve logging for SQS in Kibana and add maat_reference

alexdesi:
## What

This PR adds support for additional params for the SQS logging on Kibana.
Log maat_reference to the log on publish messages.

[Link to story](https://dsdmoj.atlassian.net/browse/LASB2037)

### DIFF
--- a/app/services/sqs/message_publisher.rb
+++ b/app/services/sqs/message_publisher.rb
@@ -2,15 +2,16 @@
 
 module Sqs
   class MessagePublisher < ApplicationService
-    def initialize(message:, queue_url: nil, sqs_client: Aws::SQS::Client.new)
+    def initialize(message:, queue_url: nil, log_info: {}, sqs_client: Aws::SQS::Client.new)
       @sqs_client = sqs_client
       @queue_url = queue_url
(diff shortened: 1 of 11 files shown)


app/services/sqs/message_publisher.rb
```diff
@@ -21,6 +22,14 @@ def messaging_enabled?
       queue_url.present?
     end
 
-    attr_reader :sqs_client, :message, :queue_url
+    def log_message
+      log_info.merge(
+        request_id: Current.request_id,
+        queue: queue_url,
+      ).map { |k, v| "#{k}: #{v}" }
+       .join(", ")
```
alexdesi, line 30: It takes the hash `log_info` and merges additional key-value pairs into it. It then transforms the result into a string representation of each pair and then joins all the strings together with commas.  The resulting string is used for the logging.

alexdesi:
This is successfully tested. We can search for `maat ref` in the Kibana logs and see also the SQS messages relevant to the meat ref. That makes debugging much more straightforward.

Screenshot Example on Staging: 
Screenshot 2023-06-12 at 09 54 29

nittagr (approved):

PR #1150, opened 2023-06-14
LASB-2116: Verify that defendant_id is linked before of a unlinkinking

alexdesi:
## What

in VCD, when you unlink an unlinked defendant, you get success,
but it generates a background error, and Sidekiq retries it several times (for days).

[Link to story](https://dsdmoj.atlassian.net/browse/2116)

### DIFF
--- a/app/controllers/api/internal/v2/laa_references_controller.rb
+++ b/app/controllers/api/internal/v2/laa_references_controller.rb
@@ -20,6 +20,7 @@ module Api
 
         def update
           contract = UnlinkDefendantContract.new.call(**transformed_params)
+
           if contract.success?
             enqueue_unlink
             head :accepted
@@ -37,7 +38,13 @@ module Api
           end
         end
 
+        def laa_reference
+          @laa_reference ||= LaaReference.find_by(defendant_id: transformed_params[:defendant_id])
+        end
+
         def enqueue_unlink
+          check_defendant_presence!
+
           UnlinkLaaReferenceWorker.perform_async(
             Current.request_id,
             transformed_params[:defendant_id],
@@ -45,6 +52,15 @@ module Api
             transformed_params[:unlink_reason_code],
             transformed_params[:unlink_other_reason_text],
           )
+        end
+
+        def check_defendant_presence!
+          defendant_id = transformed_params[:defendant_id]
+          laa_reference = LaaReference.find_by(defendant_id: defendant_id)
+
+          if laa_reference.nil?
+            raise ActiveRecord::RecordNotFound, "Defendant '#{defendant_id}' not found!"
+          end
         end
 
         def transformed_params
--- a/app/controllers/api/internal/v2/laa_references_controller.rb
+++ b/app/controllers/api/internal/v2/laa_references_controller.rb
@@ -38,10 +38,6 @@ module Api
           end
         end
 
-        def laa_reference
-          @laa_reference ||= LaaReference.find_by(defendant_id: transformed_params[:defendant_id])
-        end
-
         def enqueue_unlink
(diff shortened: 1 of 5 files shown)


app/controllers/api/internal/v2/laa_references_controller.rb
```diff
@@ -37,6 +39,17 @@ def enforce_contract!(contract)
...
         end
 
+        def check_defendant_presence!
+          defendant_id = transformed_params[:defendant_id]
+          laa_reference = LaaReference.find_by(defendant_id: defendant_id)
+
+          if laa_reference.nil?
+            raise ActiveRecord::RecordNotFound, "Defendant NOT FOUND: defendant_id:'#{defendant_id}'"
+          elsif laa_reference.linked == false
+            raise ActiveRecord::RecordNotFound, "Defendant is already non-linked: defendant_id:'#{defendant_id}'"
```
Obsiye, line 49: instead of `non-linked` could `unlinked` be used in the message ? Only so that it follows the current standard.

alexdesi: Yes, probably it's better to use "unlinked".

app/controllers/api/internal/v2/laa_references_controller.rb
```diff
@@ -47,6 +50,15 @@ def enqueue_unlink
           )
         end
 
+        def check_defendant_presence!
+          defendant_id = transformed_params[:defendant_id]
+          laa_reference = LaaReference.find_by(defendant_id: defendant_id)
+
+          if laa_reference.nil?
+            raise ActiveRecord::RecordNotFound, "Defendant '#{defendant_id}' not found!"
```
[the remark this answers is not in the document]
alexdesi: @Obsiye, yes, let's make the errors consistent. I'll fix the error format.

app/services/laa_reference_unlinker.rb
```diff
@@ -6,7 +6,12 @@ def initialize(defendant_id:, user_name:, unlink_reason_code:, unlink_other_reas
     @user_name = user_name
     @unlink_reason_code = unlink_reason_code
     @unlink_other_reason_text = unlink_other_reason_text
-    @laa_reference = LaaReference.find_by!(defendant_id: defendant_id, linked: true)
+
+    @laa_reference = LaaReference.find_by(defendant_id: defendant_id)
+
+    unless laa_reference.linked
+      Rails.logger.warn("LaaReferenceUnlinker: the defendant '#{defendant_id}' is already unlinked.")
```
Obsiye, line 13: Hi, just to confirm that this will not be sent back to CDUI? The unlinked error

alexdesi: that's right, it is only logged to Kibana.

app/controllers/api/internal/v2/laa_references_controller.rb
```diff
@@ -47,6 +50,14 @@ def enqueue_unlink
           )
         end
 
+        def check_defendant_presence!
+          laa_reference = LaaReference.find_by(defendant_id: transformed_params[:defendant_id])
```
alexdesi, line 54: `find_by` does **not** trigger an exception

`find_by1` does, but we want to customise the message (line #57), so we don't use it.

app/services/laa_reference_unlinker.rb
alexdesi, line 13: Here we just log, we don't raise exceptions if already unlinked.
By doing so we can fix potential not-in-sync data

Obsiye (approved):

mtac50 (approved):

PR #1163, opened 2023-06-28
LASB-2170: Fix search for LaaReference in unlink.

alexdesi:
## What
The error is explained here
[Link to story](https://dsdmoj.atlassian.net/browse/LASB-2170:)

Now the LaaReference search only for defendant_id thatare linked.

### DIFF
--- a/app/services/laa_reference_unlinker.rb
+++ b/app/services/laa_reference_unlinker.rb
@@ -7,14 +7,12 @@ class LaaReferenceUnlinker < ApplicationService
     @unlink_reason_code = unlink_reason_code
     @unlink_other_reason_text = unlink_other_reason_text
 
-    @laa_reference = LaaReference.find_by(defendant_id: defendant_id)
-
-    unless laa_reference.linked
-      Rails.logger.warn("LaaReferenceUnlinker: the defendant '#{defendant_id}' is already unlinked.")
-    end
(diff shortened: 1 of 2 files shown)


app/controllers/api/internal/v1/defendants_controller.rb
```diff
@@ -36,6 +36,14 @@ def enqueue_unlink
           )
         end
 
+        def check_defendant_presence!
```
alexdesi, line 39: Call `check_defendant_presence!` before `UnlinkLaaReferenceWorker.perform_async`
This is to avoid having the error inside `UnlinkLaaReferenceWorker`

app/services/laa_reference_unlinker.rb
```diff
@@ -7,14 +7,12 @@ def initialize(defendant_id:, user_name:, unlink_reason_code:, unlink_other_reas
...
-    @laa_reference = LaaReference.find_by(defendant_id: defendant_id)
-
-    unless laa_reference.linked
-      Rails.logger.warn("LaaReferenceUnlinker: the defendant '#{defendant_id}' is already unlinked.")
-    end
+    @laa_reference = LaaReference.find_by(defendant_id: defendant_id, linked: true)
   end
 
   def call
+    return unless laa_reference
```
alexdesi, line 14: it returns `nil`

PR #1217, opened 2024-02-02
[LASB-2955] Blank defendant logging

JHunt16:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LASB-2955)

Log an error and Sentry message when a blank defendant is received in the prosecution_case_controller.

## Checklist

### DIFF
--- a/app/services/common_platform/api/search_prosecution_case.rb
+++ b/app/services/common_platform/api/search_prosecution_case.rb
@@ -51,9 +51,9 @@ module CommonPlatform
       def log_error_for_blank_defendants
         if array_not_empty(@blank_defendants)
           @blank_defendants.each do |blank_defendant|
-            Rails.logger.error("The defendant with the defendantId [" + blank_defendant + "] is blank (missing defendantFirstName, defendantLastName, defendantDOB, defendantNINO and hearingSummary)")
+            Rails.logger.error("The defendant with the defendantId [#{blank_defendant}] is blank (missing defendantFirstName, defendantLastName, defendantDOB, defendantNINO and hearingSummary)")
             # Send an alert to sentry
-            Sentry.capture_message("The defendant with the defendantId [" + blank_defendant + "] is blank (missing defendantFirstName, defendantLastName, defendantDOB, defendantNINO and hearingSummary)")
+            Sentry.capture_message("The defendant with the defendantId [#{blank_defendant}] is blank (missing defendantFirstName, defendantLastName, defendantDOB, defendantNINO and hearingSummary)")
           end
         end
       end
--- a/app/services/common_platform/api/search_prosecution_case.rb
+++ b/app/services/common_platform/api/search_prosecution_case.rb
@@ -28,13 +28,13 @@ module CommonPlatform
               end
             end
 
-            check_hearing_summaries_for_blank_defendants(prosecution_case["hearingSummary"])
+            remove_defendant_with_hearing_summary_from_blank_defendants(prosecution_case["hearingSummary"])
             log_error_for_blank_defendants
           end
         end
       end
 
-      def check_hearing_summaries_for_blank_defendants(hearing_summaries)
+      def remove_defendant_with_hearing_summary_from_blank_defendants(hearing_summaries)
         if array_not_empty(hearing_summaries)
           hearing_summaries.each do |hearing_summary|
             hearing_summary["defendantIds"].each do |defendant_id|
--- a/app/services/common_platform/api/search_prosecution_case.rb
+++ b/app/services/common_platform/api/search_prosecution_case.rb
@@ -7,6 +7,7 @@ module CommonPlatform
 
       def initialize(params)
         @response = ProsecutionCaseSearcher.call(**params)
+        @blank_defendants = []
       end
 
       def call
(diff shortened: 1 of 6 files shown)


app/services/common_platform/api/search_prosecution_case.rb
```diff
@@ -11,10 +11,51 @@ def initialize(params)
...
+        check_for_blank_defendants(prosecution_cases)
         record_prosecution_cases
       end
 
+      def check_for_blank_defendants(prosecution_cases)
+        if !prosecution_cases.nil? && !prosecution_cases.empty?
+          prosecution_cases.each do |prosecution_case|
+            if !prosecution_case["defendantSummary"].nil? && !prosecution_case["defendantSummary"].empty?
+              prosecution_case["defendantSummary"].each do |defendant|
+                # If the defendant is missing these values in their summary
```
lkm0287, line 23: Rather than commenting what these complex conditions are doing, it is typical in development to pull them out in to their own method with the method name explaining what it is doing. This makes it easier to understand and test. Here the above condition could be a method has_defendant_summary_and_prosecution_case or something?

lkm0287: It also makes the comment not necessary.

app/services/common_platform/api/search_prosecution_case.rb
```diff
@@ -11,10 +11,51 @@ def initialize(params)
...
+                    @empty_defendants = []
+                  end
+                  # Add this defendant to an array of empty defendants (we need to also verify they don't have a hearing summary further down)
+                  @empty_defendants.push(defendant["defendantId"])
+                end
+              end
+            end
+
+            # We need to check the hearing summaries for the blank defendants
+            if !prosecution_case["hearingSummary"].nil? && !prosecution_case["hearingSummary"].empty?
```
lkm0287, line 35: Each of these blocks of code could be its own method too.

lkm0287: Basically we want the code readable enough that we don't need the comments.

spec/fixtures/files/prosecution_case_search_result_with_only_blank_defendant.json
```diff
@@ -0,0 +1,135 @@
...
+      "defendantSummary": [
+        {
+          "proceedingsConcluded": true,
+          "representationOrder": {
+            "applicationReference": "LAA-20191601",
+            "effectiveFromDate": "2019-09-12",
+            "effectiveToDate": "2019-12-12",
+            "laaContractNumber": "99998888"
+          },
+          "defendantMiddleName": "Andrew",
```
lkm0287, line 17: Are we also publicly leaking real prod data here? 😬

JHunt16: Hopefully not... I copied it from an existing file in the same directory - It has a clearly fake defendantASN (ARREST123), but other parts may not be. I will raise this with the team next week as it's an interesting issue.

JHunt16: If there was any issue with this data, then we probably need to do some sort of audit as there is at least one existing file with the same data.

app/services/common_platform/api/search_prosecution_case.rb
JHunt16: Makes sense. I've split things up into separate methods where appropriate. Let me know if you have any more feedback on this 😄

app/services/common_platform/api/search_prosecution_case.rb
JHunt16: I've split into a single method to check if the array is empty, as there are multiple checks here for the same sort of thing. Hopefully it is more readable than before (and reduces some duplication).

app/services/common_platform/api/search_prosecution_case.rb
```diff
@@ -7,14 +7,65 @@ class SearchProsecutionCase < ApplicationService
...
       end
 
       def call
         check_response_status
-
+        check_for_blank_defendants(prosecution_cases)
         record_prosecution_cases
       end
 
+      def check_for_blank_defendants(prosecution_cases)
```
lkm0287, line 19: I'd name this logBlankDefendants or something because that's what it's ultimately doing rather than like a check (returning a boolean).

app/services/common_platform/api/search_prosecution_case.rb
```diff
@@ -7,14 +7,65 @@ class SearchProsecutionCase < ApplicationService
...
+              end
+            end
+
+            check_hearing_summaries_for_blank_defendants(prosecution_case["hearingSummary"])
+            log_error_for_blank_defendants
+          end
+        end
+      end
+
+      def check_hearing_summaries_for_blank_defendants(hearing_summaries)
```
lkm0287, line 37: I'd call this removeDefendantWithHearingFromBlankDefendants or something?

lkm0287 (dismissed):

app/services/common_platform/api/search_prosecution_case.rb
JHunt16: I've opted to keep this one as-is, as within this method I already have a call to an existing method: log_error_for_blank_defendants, which I think covers it.

lkm0287 (approved):

PR #1233, opened 2024-06-25
Acd 171 compare v1 and v2 data

mtac50:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-171)

- [x] Create a rake task for perform data comparisons between the V! and V@ defendant endpoint

## Checklist


exonian (approved):
Haven't parsed this in full detail but nothing stands out as a problem and if it does what you need for disposable code 👍

PR #1281, opened 2025-01-22
ACD-433: Add applicationSummary on SearchProsecutionCase.

alexdesi:
Jira: https://dsdmoj.atlassian.net/browse/ACD-433

## What
ACD-433: Add application summary to v2/prosecution_cases endpoint

      - Refactored spec/models/prosecution_case_spec.rb to use WebMock for improved transparency and maintainability.
      - Rename prosecution_case_searcher -> prosecution_case_fetcher, to avoid confusion and adopt an existing naming convention.
      - add new fixtures spec/fixtures/files/prosecution_cases_v2.json
      - Note: ProsecutionCase makes API calls, which should be moved to services. A larger refactor is needed.

### DIFF
--- a/app/models/hmcts_common_platform/application_summary.rb
+++ b/app/models/hmcts_common_platform/application_summary.rb
@@ -1,0 +1,40 @@
+module HmctsCommonPlatform
+  class ApplicationSummary
+    attr_accessor :data
+
+    def initialize(data)
+      @data = HashWithIndifferentAccess.new(data || {})
+    end
+
+    def id
+      data[:applicationId]
+    end
+
+    def reference
+      data[:applicationReference]
+    end
+
+    def title
+      data[:applicationTitle]
+    end
+
+    def received_date
+      data[:receivedDate]
+    end
+
+    def to_json(*_args)
+      to_builder.attributes!
+    end
+
+  private
+
+    def to_builder
+      Jbuilder.new do |application_summary|
+        application_summary.id id
+        application_summary.reference reference
+        application_summary.title title
+        application_summary.received_date received_date
+      end
+    end
+  end
+end
--- a/app/models/hmcts_common_platform/prosecution_case_summary.rb
+++ b/app/models/hmcts_common_platform/prosecution_case_summary.rb
@@ -26,6 +26,12 @@ module HmctsCommonPlatform
       end
     end
 
+    def application_summaries
+      Array(data[:applicationSummary]).map do |application_summary_data|
+        HmctsCommonPlatform::ApplicationSummary.new(application_summary_data)
+      end
+    end
+
     def to_json(*_args)
       to_builder.attributes!
     end
@@ -38,6 +44,7 @@ module HmctsCommonPlatform
         case_summary.case_status case_status
         case_summary.defendant_summaries defendant_summaries.map(&:to_json)
         case_summary.hearing_summaries hearing_summaries.map(&:to_json)
+        case_summary.application_summaries application_summaries.map(&:to_json)
       end
     end
   end
--- a/app/services/common_platform/api/prosecution_case_fetcher.rb
+++ b/app/services/common_platform/api/prosecution_case_fetcher.rb
@@ -2,7 +2,7 @@
 
 module CommonPlatform
   module Api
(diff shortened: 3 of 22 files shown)


mo-zag (dismissed):

app/services/common_platform/api/prosecution_case_fetcher.rb
```diff
@@ -2,7 +2,7 @@
 
 module CommonPlatform
   module Api
-    class ProsecutionCaseSearcher < ApplicationService
+    class ProsecutionCaseFetcher < ApplicationService
```
alexdesi, line 5: Renamed to avoid confusion with `SearchProsecutionCase` and to adopt an existing naming convention.

app/models/hmcts_common_platform/prosecution_case_summary.rb
```diff
@@ -26,6 +26,12 @@ def hearing_summaries
       end
     end
 
+    def application_summaries
```
alexdesi, line 29: this should be singular `ApplicationSummary`

lib/schemas/global/apiCourtsDefinitions.json
```diff
@@ -77,7 +77,14 @@
...
-      "type": "string"
+      "type": "string",
+      "enum": [
+        "DRAFT",
+        "LISTED",
+        "FINALISED",
+        "EJECTED",
+        "IN_PROGRESS"
+      ]
     },
```
alexdesi, line 88: I am checking this change ...

app/models/hmcts_common_platform/prosecution_case_summary.rb
alexdesi: I've reconsidered my previous comment.
Since application_summaries returns an array, it makes sense for the name to be plural.

Even though the property inside the JSON is singular (applicationSummary), it's consistent with how defendant_summaries and hearing_summaries are handled.

So, I'd suggest we stick with the existing convention for now.

patrick-laa (dismissed):

app/models/hmcts_common_platform/application_summary.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+  private
+
+    def to_builder
+      Jbuilder.new do |application_summary|
+        application_summary.id id
+        application_summary.reference reference
+        application_summary.title title
+        application_summary.receivedDate received_date
```
patrick-laa, line 36: Should this be `received_date`? I think all other parts of the payload are underscored

spec/models/hmcts_common_platform/application_summary_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:hearing_summary) { described_class.new(data) }
+
+  context "with all fields" do
+    let(:data) { JSON.parse(file_fixture("application_summary/all_fields.json").read) }
+
+    it "matches the HMCTS Common Platform schema" do
+      expect(data).to match_json_schema(:application_summary)
+    end
+
+    it { expect(hearing_summary.id).to eql("6cd6494e-5409-420a-bd0b-f589dbb2466b") }
```
patrick-laa, line 11: Maybe it's worth adding a spec that verifies the output of `summary.to_json`

spec/models/hmcts_common_platform/application_summary_spec.rb
```diff
@@ -0,0 +1,16 @@
+RSpec.describe HmctsCommonPlatform::ApplicationSummary, type: :model do
+  let(:hearing_summary) { described_class.new(data) }
```
patrick-laa, line 2: I think this should be `application_summary`, not `hearing_summary`

app/models/hmcts_common_platform/application_summary.rb
alexdesi: yes, that's right! thanks!

spec/models/hmcts_common_platform/application_summary_spec.rb
alexdesi: yes! that would allow spot the error with `receivedDate` instead of `received_date`
Done!

patrick-laa (approved):

PR #1289, opened 2025-03-03
ACD-587: Update to Rails 7.2

alexdesi:
JIRA: https://dsdmoj.atlassian.net/browse/ACD-587

## What
Update Rails 7 -> 7.2
Fix various Warnings due to old dependencies.

## Checklist


app/services/maat_api/maat_reference_validator.rb
```diff
@@ -10,7 +10,9 @@ def initialize(maat_reference:, connection: MaatApi::Connection.call)
     end
 
     def call
-      connection.post(URL, { maatId: maat_reference, caseUrn: "XYZ" }) if connection.present?
+      raise "Connection is blank" if connection.blank?
```
patrick-laa, line 13: This is a bit annoying. In the local dev environment, connection is blank, and it is helpful for this not to raise an error, because it allows you to access UI flows that are otherwise impossible (https://mojdt.slack.com/archives/C0832R7RHTN/p1741598387938079?thread_ts=1741342371.934119&cid=C0832R7RHTN).

Is it possible to keep the behaviour the same as it was here?

PR #1312, opened 2025-03-18
ACD-618: Deal with situation where defendant ID spans multiple cases

patrick-laa:
## What

Bug outline:

VCD user looks up case `CaseC` featuring defendant B, who has defendantID `B1` and masterDefendantId `B2`. ID `B1` gets sent to VCD as part of the `CaseC` payload. When the user clicks on defendant B, VCD makes a request for details to CDA using ID `B1`.

CDA then looks up the first local prosecution case record linked locally to ID `B1`. But sometimes this is not case `CaseC` but rather case `CaseD`. When CDA then pulls defendant details for `CaseD` from HMCTS, defendant B is listed as having ID `B2`, not `B1`. This means that, searching in the context of CaseB, no details associated with ID B1 are found. So CDA returns a 404, which results in a `JsonApiClient::Errors::NotFound` being recorded [to Sentry](https://ministryofjustice.sentry.io/issues/2066105222/?project=5428270&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=4) by VCD

Because, when we search for defendant details, we don't send the case identifier from VCD to CDA, when there are multiple local cases associated with a defendant ID in CDA we can't specify which local case to look up. BUT we can effectively work around this issue by iterating through them until we find one associated with an HMCTS case that lists the defendant ID we have.

There is some suggestion that defendant IDs shouldn't be duplicated across cases. However, given they _are_ duplicated, and this issue affects around 10 VCD interactions a day, working around this weirdness of the data is more pragmatic than ignoring it.

### DIFF
--- a/app/services/common_platform/api/defendant_finder.rb
+++ b/app/services/common_platform/api/defendant_finder.rb
@@ -5,7 +5,6 @@ module CommonPlatform
     class DefendantFinder < ApplicationService
       def initialize(defendant_id:)
         @defendant_id = defendant_id
-        @prosecution_case_defendant = ProsecutionCaseDefendantOffence.find_by(defendant_id:)
       end
 
       def call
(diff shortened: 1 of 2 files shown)


app/services/common_platform/api/defendant_finder.rb
```diff
@@ -15,30 +14,50 @@ def call
...
-        @common_platform_defendant ||= common_platform_prosecution_case&.defendants&.find { |d| d.id.eql?(defendant_id) }
+        return @common_platform_defendant if @common_platform_defendant
+        return unless prosecution_case_urns
+
+        prosecution_case_urns.each do |urn|
+          prosecution_case = common_platform_prosecution_case(urn)
+
+          next unless prosecution_case
+
+          @common_platform_defendant ||= prosecution_case&.defendants&.find { |d| d.id.eql?(defendant_id) }
```
alexdesi, line 25: The &. before .find is unnecessary because line 23

patrick-laa: Good spot! `prosecution_case` can't be nil but I suppose `defendants` still could be technically, so we can lose the first `&` but not the second one.

app/services/common_platform/api/defendant_finder.rb
```diff
@@ -15,30 +14,50 @@ def call
...
+
+          break if @common_platform_defendant
+        end
+
+        @common_platform_defendant
```
alexdesi, line 30: Minor comment:
what about return nil here (instead of relying on instance variable default)
and direct return
`return @common_platform_defendant if @common_platform_defendant`
on line 27?
I think it's more clear: once you find a @common_platform_defendant you return!

patrick-laa: Nice! That's much clearer.

PR #1313, opened 2025-03-18
Only run test-branch on branches

patrick-laa:
## What

`test-branch` is currently running after we merge to main, even though `test-build-deploy-main-to-live-cluster` also runs, which does exactly the same things. This looks to me unintentional, and I think it's an indentation error in the directive that tells `test-branch` _not_ to run on merge to main.


alexdesi (approved):
✅

PR #1314, opened 2025-03-18
ACD-836 (better version): Read canonical MAAT directly, do not infer from offences

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-615)

Often the offences HMCTS give us list inaccurate MAATs, and we have been relying on them to tell us the actual MAAT. However, the MAAT is actually available in our own database, because we track the authoritative MAAT ID in our LaaReference table. This record represents what caseworkers have told us the MAAT ID is, so is a valid value to display in VCD to caseworkers regardless of whether HMCTS's own records have subsequently been corrupted.


app/models/defendant.rb
```diff
@@ -98,11 +98,7 @@ def valid_maat_reference?
   end
 
   def _maat_reference
-    refs = offences.map(&:maat_reference).uniq.compact
-
-    raise(Errors::DefendantError, "Too many maat references: #{refs}") if refs.size > 1
-
-    refs&.first
+    LaaReference.find_by(defendant_id: id, linked: true)&.maat_reference
```
patrick-laa, line 101: Note that this doesn't help with the v2 case summary endpoint, which doesn't pass back this field and instead just passes back the offence summaries, and CDAPI pulls out the first of these and uses it as "the" MAAT number. So we would need to change CDAPI to use this canonical MAAT reference when calling that endpoint otherwise the MAAT ID could appear differently on the case summary page vs the defendant page.

alexdesi (approved):
it looks good ✅

PR #1325, opened 2025-03-24
Add Rubocop to CI

patrick-laa:
At the moment we're not catching linter errors in CI


patrick-laa:
Never mind, turns out it is in there

PR #1327, opened 2025-03-25
ACD-623: Move around what points to what

patrick-laa:
## What
- Point dev to our mock
- Point test and staging to their dev/mock

https://dsdmoj.atlassian.net/browse/ACD-623


alexdesi:
@patrick-laa if I am not wrong, @muthusubra suggested this: 
 
`VCD UI STAGING -> CD API STAGING → CDA TEST -> Common Platform DEV`

(you changed CDA staging -> CP Dev.  https://apigw.dev.cjscp.org.uk/LAA/v1)

It sounds confusing, I think the point here is that we are going to get rid of staging soon, and VCD-staging will turn to VCD-test.

Should we postpone this changes until we make those env changes? (Remove staging, Add VCD test)

alexdesi (approved):
✅

PR #1329, opened 2025-03-27
ACD-574: Record court application body locally

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-574)

- When a court application is retrieved, store it in the `prosecution_cases` table
- For each offence, store a row in the `prosecution_cases_defendant_offences` table along with the application type (in a new column)
- Add a warning to the code acknowledging that a "ProsecutionCase" may not actually be a prosecution case.

## Checklist

### DIFF
--- a/app/controllers/api/internal/v2/court_applications_controller.rb
+++ b/app/controllers/api/internal/v2/court_applications_controller.rb
@@ -12,9 +12,9 @@ module Api
 
           case response_data.status
           when 200
-            model = HmctsCommonPlatform::CourtApplicationSummary.new(response_body)
-            CourtApplicationRecorder.call(params[:id], model)
-            render json: model.to_json, status: :ok
+            application_summary = HmctsCommonPlatform::CourtApplicationSummary.new(response_body)
+            CourtApplicationRecorder.call(params[:id], application_summary)
+            render json: application_summary.to_json, status: :ok
           when 404
             head :not_found
           else
--- a/app/controllers/api/internal/v2/court_applications_controller.rb
+++ b/app/controllers/api/internal/v2/court_applications_controller.rb
@@ -12,7 +12,9 @@ module Api
 
           case response_data.status
           when 200
-            render json: HmctsCommonPlatform::CourtApplicationSummary.new(response_body).to_json, status: :ok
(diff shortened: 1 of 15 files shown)


app/controllers/api/internal/v2/court_applications_controller.rb
```diff
@@ -12,7 +12,9 @@ def show
 
           case response_data.status
           when 200
-            render json: HmctsCommonPlatform::CourtApplicationSummary.new(response_body).to_json, status: :ok
+            model = HmctsCommonPlatform::CourtApplicationSummary.new(response_body)
```
alexdesi, line 15: `model` sounds a bit generic, what about
`court_application_summary` or `application_summary` ?

app/models/prosecution_case.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# WARNING! Despite the name, a ProsecutionCase DB record can contain _either_
+# a prosecution case our a court application record
```
alexdesi, line 4: 👍 for the comment!

that's a good point, I was wondering if we can clarify the distintion in the code,
maybe adding an abstraction, called `LegalCase` or something,
and 
`ProsecutionCase < LegalCase`

`CourtApplication < LegalCase`

and we can associate LegalCase with the existing `prosecution_cases` table

could that help? 🤔

db/structure.sql
```diff
@@ -198,7 +198,8 @@ CREATE TABLE public.prosecution_case_defendant_offences (
     status_date timestamp without time zone,
     effective_start_date timestamp without time zone,
     effective_end_date timestamp without time zone,
-    defence_organisation json
+    defence_organisation json,
+    application_type character varying
```
alexdesi, line 202: Just asking should we put some limits on this, for example: 
`character varying(255)`

alexdesi (approved):
It looks good,!
tests are very clear 👍 
added minor comments ✅

app/models/prosecution_case.rb
patrick-laa: Great idea! I have done exactly that.

PR #1330, opened 2025-03-28
ACD-437: Accept new application concluded payload to prosecution conclusion endpoint

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-437)

When a court application is concluded the payload from HMCTS gets sent to the existing prosecution_concluded endpoint. Currently only the v1 endpoint is used. A v2 endpoint exists but is a copy-paste of the code of v1.

- Turn controller logic into a concern that is shared by both v1 and v2 so we can update logic in both at the same time
- Change param whitelisting and contract enforcement logic to accept new params and allow old params to be missing
- Retrieve LAA reference using subjectId if defendantId is not available
- Send the message on to MAAT

## Checklist

### DIFF
--- a/app/controllers/api/external/v1/prosecution_conclusions_controller.rb
+++ b/app/controllers/api/external/v1/prosecution_conclusions_controller.rb
@@ -2,75 +2,7 @@ module Api
   module External
     module V1
       class ProsecutionConclusionsController < ApplicationController
-        def create
-          enforce_contract!
-
-          prosecution_conclusion_params["prosecutionConcluded"].each do |pc|
-            laa_reference = LaaReference.find_by(defendant_id: pc["defendantId"], linked: true)
-
-            next unless laa_reference
-
-            Sqs::MessagePublisher.call(
-              message: pc.to_h.merge("maatId" => laa_reference.maat_reference),
-              queue_url: Rails.configuration.x.aws.sqs_url_prosecution_concluded,
-              log_info: { maat_reference: laa_reference.maat_reference },
-            )
-          end
-
-          head :accepted
-        end
-
-      private
-
-        def enforce_contract!
-          unless contract.success?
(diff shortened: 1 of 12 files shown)


app/controllers/concerns/prosecution_concludable.rb
```diff
@@ -0,0 +1,83 @@
...
+      message = "Prosecution conclusion contract failed with: #{contract.errors.to_hash}"
+      raise Errors::ContractError, message
+    end
+  end
+
+  def contract
+    ProsecutionConclusionContract.new.call(prosecution_conclusion_params.to_hash)
+  end
+
+  def retrieve_laa_reference(param_set)
```
alexdesi, line 35: to be more explicit maybe name it to`def get_linked_laa_reference`  ?

app/controllers/api/external/v1/prosecution_conclusions_controller.rb
```diff
@@ -2,75 +2,7 @@ module Api
...
-                        ],
-                      },
-                    ],
-                  },
-                ],
-              },
-            ],
-          )
-        end
+        include ProsecutionConcludable
```
alexdesi, line 5: that's a nice Dryfication :)
but if I am not wrong the V1 is not used anymore ...
well in anycase we can remove refactor it later

alexdesi (approved):
it looks good,
just minor comments.
✅

app/controllers/api/external/v1/prosecution_conclusions_controller.rb
patrick-laa: Annoyingly _only_ v1 is used! E.g. our mock uses https://github.com/ministryofjustice/hmcts-common-platform-mock-api/blob/main/app/services/prosecution_conclusion_publisher.rb#L4. Ganga said that because we intend to migrate to v2 eventually, we should make the change to both endpoints.

PR #1331, opened 2025-03-31
Small refactor: remove alias + Comments

alexdesi:
## What
to make it more explicit the calls to /LAA/v1/hearing/results (CP) which are obscured by an alias.


patrick-laa (dismissed):
Yes! This makes it MUCH clearer what's going on, so that we're no longer hiding all the extra calls we make to common platform.

patrick-laa (approved):

PR #1333, opened 2025-04-03
ACD-647 and ACD-636: Short ID and subject summary are in both application payloads

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-647)
[Link to story](https://dsdmoj.atlassian.net/browse/ACD-636)

Pass on to client the application short ID and the subject summary in both application payloads - the application summary in the prosecution cases search endpoint and the 'get court application' endpoint.

## Checklist

### DIFF
--- a/lib/schemas/global/search/apiApplicationSummary.json
+++ b/lib/schemas/global/search/apiApplicationSummary.json
@@ -19,6 +19,156 @@
     "receivedDate": {
       "description": "The date the application was received",
       "$ref": "http://justice.gov.uk/core/courts/external/apiCourtsDefinitions.json#/definitions/datePattern"
+    },
+    "applicationShortId": {
+      "description": "An HMCTS-generated unique string identifying the application",
+      "type": "string"
+    },
+    "subjectSummary":  {
+      "id": "http://justice.gov.uk/progression/query/laa/subject-summary.json",
+      "type": "object",
+      "properties": {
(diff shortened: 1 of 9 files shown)


lib/schemas/global/search/apiApplicationSummary.json
```diff
@@ -1,24 +1,174 @@
...
-      "description": "The title of the application",
-      "type": "string"
-    },
-    "receivedDate": {
-      "description": "The date the application was received",
-      "$ref": "http://justice.gov.uk/core/courts/external/apiCourtsDefinitions.json#/definitions/datePattern"
-    }
-  }
-}
+{
```
alexdesi, line 1: It would be great only to have the changes in the diff, instead of the entire file.
What is it? end of line or maybe different indentation? 🤔

patrick-laa: I wish I knew what was up with these schema files! I can't save the file inside VScode without Git thinking I've replaced the whole contents!

alexdesi: I think I had to paste/copy it inside sublime text and then past again into VScode,
it's okay, the only reason it that would make review easier, but it's not worth the time.

alexdesi (approved):
nice one! 🕺  ✅

PR #1334, opened 2025-04-04
ACD-434: Court Application LAA references

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-434)

Build new `/api/internal/v2/court_application_laa_references` create and update endpoints, which mirror the existing `/api/internal/v2/laa_references` endpoints, doing the same thing but for court applications instead of prosecution cases.

Annoyingly, even though the code is pretty similar, the difference between subject_id/defendant_id throughout and CourtApplicationDefendantOffence/ProsecutionCaseDefendantOffence makes it harder to reuse code without breaking everything up into little chunks that have to be composed or inherited together, making everything harder to read.

So I've chosen not to be particularly DRY here, opting to create new classes that look quite similar to existing ones, for the sake of readability.

I've tested integration with the mock Common Platform API locally and everything works there, but I don't have an easy way of testing this plays nicely with the MAAT API, so have had to rely on specs that ensure we send sane-looking payloads to MAAT.

## Checklist

### DIFF
--- a/app/contracts/new_court_application_laa_reference_contract.rb
+++ b/app/contracts/new_court_application_laa_reference_contract.rb
@@ -1,7 +1,7 @@
 class NewCourtApplicationLaaReferenceContract < Dry::Validation::Contract
   option :uuid_validator, default: -> { CommonPlatform::UuidValidator }
   option :maat_reference_validator, default: -> { MaatApi::MaatReferenceValidator }
-  option :link_validator, default: -> { CouertApplicationLinkValidator }
+  option :link_validator, default: -> { CourtApplicationLinkValidator }
 
   params do
     optional(:maat_reference).value(:integer, lt?: 999_999_999)
--- a/app/contracts/new_court_application_laa_reference_contract.rb
+++ b/app/contracts/new_court_application_laa_reference_contract.rb
@@ -1,0 +1,23 @@
+class NewCourtApplicationLaaReferenceContract < Dry::Validation::Contract
+  option :uuid_validator, default: -> { CommonPlatform::UuidValidator }
+  option :maat_reference_validator, default: -> { MaatApi::MaatReferenceValidator }
+  option :link_validator, default: -> { CouertApplicationLinkValidator }
+
+  params do
+    optional(:maat_reference).value(:integer, lt?: 999_999_999)
+    required(:user_name).value(:string, min_size?: 1, max_size?: 10)
+    required(:subject_id).value(:string)
+  end
+
+  rule(:subject_id) do
+    key.failure("is not a valid uuid") unless uuid_validator.call(uuid: value)
+    unless link_validator.call(subject_id: value)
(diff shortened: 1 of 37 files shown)


app/services/court_application_maat_link_creator.rb
```diff
@@ -0,0 +1,89 @@
...
+      message: MaatApi::LaaReferenceMessage.new(maat_api_laa_reference).generate,
+      queue_url: Rails.configuration.x.aws.sqs_url_link,
+      log_info: { maat_reference: laa_reference.maat_reference },
+    )
+  end
+
+  def post_laa_references_to_common_platform
+    offences.each do |offence|
+      post_laa_reference_to_common_platform(offence)
+    rescue StandardError => e
```
alexdesi, line 40: I think you need to wrap the `rescue` between begin/end
something like
```
offences.each do |offence|
    begin
      post_laa_reference_to_common_platform(offence)
    rescue StandardError => e
    ...
    end
  end

File: app/contracts/new_court_application_laa_reference_contract.rb [hunk 2]
```diff
@@ -0,0 +1,23 @@
+class NewCourtApplicationLaaReferenceContract < Dry::Validation::Contract
+  option :uuid_validator, default: -> { CommonPlatform::UuidValidator }
+  option :maat_reference_validator, default: -> { MaatApi::MaatReferenceValidator }
+  option :link_validator, default: -> { CourtApplicationLinkValidator }
+
+  params do
+    optional(:maat_reference).value(:integer, lt?: 999_999_999)
+    required(:user_name).value(:string, min_size?: 1, max_size?: 10)
```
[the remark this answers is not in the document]
patrick-laa: Yeah, I was surprised, but the same constraints apply to the username in VCD, and so possibly they apply upstream in the MAAT API and HMCTS.

app/services/court_application_maat_link_creator.rb
patrick-laa: I tried that, and Rubocop gives me a `Redundant `begin` block detected.` error - turns out you don't need it.

alexdesi: yes, you're right, I've just checked.
the block was necessary before Ruby 2.5,
thanks for trying!

app/services/court_application_link_validator.rb
```diff
@@ -0,0 +1,13 @@
+class CourtApplicationLinkValidator < ApplicationService
+  def initialize(subject_id:)
+    @court_application = CourtApplicationDefendantOffence.find_by(defendant_id: subject_id)&.court_application
+  end
+
+  def call
+    (court_application&.hearing_summaries || []).any?
```
alexdesi, line 7: 👍

alexdesi (approved):
nice one 🚀 ✅

PR #1336, opened 2025-04-07
ACD-635: Speed up defendants endpoint

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-434)

Stop looking up details for hearings that are associated with the defendant's prosecution case but not with the defendant themselves. These lookups weren't actually being used to construct the response payload, so they were just wasted HTTP calls

## Checklist

### DIFF
--- a/app/models/prosecution_case.rb
+++ b/app/models/prosecution_case.rb
@@ -37,37 +37,8 @@ class ProsecutionCase < LegalCase
     body["prosecutionCaseReference"]
   end
 
-  # In order to populate "details" for a defendant (see #defendants above)
-  # we need to retrieve hearing results. If we retrieve all hearing results
-  # relating to a particular defendant, we get a full list of things like
-  # post_hearing_custody_statuses and judicial_results.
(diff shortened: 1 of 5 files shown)


app/models/prosecution_case.rb
```diff
@@ -37,6 +37,39 @@ def prosecution_case_reference
...
+      )
+      next if result.blank?
+
+      loaded << result
+      break unless load_all
+    end
+
+    @hearing_results = loaded
+  end
+
```
mo-zag, line 72: This method handles several responsibilities — would it help splitting it into smaller helper methods could improve readability and maintainability.

patrick-laa: That's fair. I need to check in with Matt, because it might be that some of this isn't going to work anyway and can be removed. Once I've spoken to him I'll refactor to simplify.

mo-zag (approved):
All good from me depending on your conversation with Matt plz let me know if you need me to review again.

PR #1338, opened 2025-04-10
Application representation order 

mo-zag:
Application representation order ticket ACD-435

## What

[ACD-435](https://dsdmoj.atlassian.net/browse/ACD-435)

The logic for handling representation orders has been kept separate to allow independent feature development. This separation ensures that new features can be added without interfering with existing functionality, providing flexibility for future enhancements.

## Checklist

[ACD-435]: https://dsdmoj.atlassian.net/browse/ACD-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/common_platform/api/court_application_representation_order_creator.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      def call
+        call_common_platform_endpoint
+      end
+
+    private
+
+      def call_common_platform_endpoint
+        offences.each do |offence|
+          case_defendant_offence = ProsecutionCaseDefendantOffence.find_by(defendant_id: subject_id, offence_id: offence[:offence_id])
```
patrick-laa, line 22: There's some nuance here, but the TL;DR is: In practice, this won't work unless you change this line to:

```ruby
application_defendant_offence = CourtApplicationDefendantOffence.find_by(....
```

The reason is that when we [save](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/main/app/services/court_application_recorder.rb#L13) court application defendant offences to the `prosecution_case_defendant_offences` table, we will always include the `application_type`. Because of the [default_scope](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/main/app/models/prosecution_case_defendant_offence.rb#L9) on ProsecutionCaseDefendantOffence, even if you use the right identifiers in your `find_by` it will still not retrieve any records that have an application_type.

In practice, this basically means that in the ruby layer we should be thinking of prosecution cases and court applications as being separate entities, even though they are stored in the same database table.

app/services/common_platform/api/court_application_record_representation_order.rb
```diff
@@ -0,0 +1,65 @@
...
+        @case_defendant_offence = case_defendant_offence
+        @offence_id = offence_id
+        @status_code = status_code
+        @application_reference = application_reference.to_s
+        @status_date = status_date
+        @effective_start_date = effective_start_date
+        @effective_end_date = effective_end_date
+        @defence_organisation = defence_organisation
+        @url = "prosecutionCases/representationOrder"\
+                "/applications/#{case_defendant_offence.prosecution_case_id}"\
```
patrick-laa, line 25: The other advantage of my previous comment is that this line could instead be:

```
"/applications/#{application_defendant_offence.court_application_id}"\
```

Which I think makes a little more sense if you don't know the weird way that the DB is structured.

app/services/common_platform/api/court_application_representation_order_creator.rb
mo-zag: This should be changed now.

app/services/common_platform/api/court_application_record_representation_order.rb
mo-zag: Updated this too thanks.

patrick-laa (approved):
Looks good to me!

PR #1340, opened 2025-04-15
Return ACD-522 application_summaries within defendant_summaries payload

mo-zag:
## What

[ACD-652](https://dsdmoj.atlassian.net/browse/ACD-652)
[ACD-522](https://dsdmoj.atlassian.net/browse/ACD-522)

Matched the defendant to an application using first and last name, and appended the corresponding application to defendant_summaries, allowing VCD to directly display the application title.

## Checklist

[ACD-652]: https://dsdmoj.atlassian.net/browse/ACD-652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ACD-522]: https://dsdmoj.atlassian.net/browse/ACD-522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/common_platform/api/search_prosecution_case.rb
```diff
@@ -20,11 +20,17 @@ def check_for_blank_defendants(prosecution_cases)
         return if prosecution_cases.nil?
 
         prosecution_cases.each do |prosecution_case|
-          prosecution_case["defendantSummary"]&.each do |defendant|
+          prosecution_case["defendantSummary"]&.map! do |defendant| # Use `map!` to update the array in place
```
patrick-laa, line 23: I don't think this is the right place to do this. The method here is called `check_for_blank_defendant` - having it actually manipulate the JSON blob is a bit suspect, particularly as a side effect of a method that is _supposed_ to just be building up an array of blank defendant IDs.

Worse, this gets called _before_ `record_prosecution_case` gets called, meaning that this is messing with our DB logs of what HMCTS is returning to us - we don't store what's returned, we store a modified version instead, which I think will bite us in the future.

Can I recommend maybe passing the application summary blobs as a separate argument to HmctsCommonPlatform::DefendantSummary when instantiating it and then filtering the blob in that model? That way we don't have to manipulate the input JSON blob itself.

spec/models/hmcts_common_platform/defendant_summary_spec.rb
```diff
@@ -1,5 +1,20 @@
...
+        "title": "Appeal at conviction here",
+        "receivedDate": "2025-04-11",
+        "subjectSummary": {
+          "defendantFirstName": "Bob",
+          "defendantLastName": "Smith",
+        },
+      },
+    ]
+  end
+  let(:defendant_summary) { described_class.new(data, :court_application_summary) }
```
patrick-laa, line 17: This is a bug - you're not passing in `court_application_summary`, you're passing in `:court_application_summary`, i.e. just a symbol. Because of that I'm guessing `defendant_summary.application_summaries` is blank, so your test "passes" but only because it's not testing anything. I think this is a sign we need better tests here.

app/models/hmcts_common_platform/defendant_court_application_summary.rb
```diff
@@ -0,0 +1,50 @@
+module HmctsCommonPlatform
+  class DefendantCourtApplicationSummary
```
patrick-laa, line 2: I still don't see the value of creating a new class here - HmctsCommonPlatform::ApplicationSummary was designed explicitly to generate a response payload for the court application summaries in the prosecution cases endpoint. It literally has no other purpose than to do what you need it to do, so why not use it?

app/models/hmcts_common_platform/defendant_summary.rb
```diff
@@ -52,12 +54,34 @@ def offence_summaries
...
+      @data[:applicationSummaries] = []
+
+      return unless court_application_summaries.is_a?(Array) && court_application_summaries.any?
+
+      court_application_summaries.each do |summary|
+        master_defendant_id = summary.dig("subjectSummary", "masterDefendantId").to_s.strip
+
+        expected_defendant_id = data[:defendantId].to_s.strip
+
+        next unless master_defendant_id == expected_defendant_id
```
patrick-laa, line 79: You still don't have any actual tests of this functionality. I'd like a test that demonstrates that if the masterDefendantId matches, the summary is included in the payload, and one that shows if it doesn't match, it's not included.

lib/schemas/global/apiDefendantCourtApplicationSummary.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
patrick-laa, line 2: This is a pain, I know, but to keep swagger up to date you're going to need to update swagger/v2/defendant_summary.json to indicate that the defendant summary is now going to include an array of application_summaries.

spec/models/hmcts_common_platform/defendant_summary_spec.rb
```diff
@@ -52,6 +54,47 @@
...
+      summaries = defendant_summary.application_summaries
+
+      expect(summaries).not_to be_empty
+      expect(summaries.first).to be_a(HmctsCommonPlatform::ApplicationSummary)
+      expect(summaries.first.data["applicationId"]).to eq("f38d0030-0b4a-4fa5-9484-bb37b1e6ab39")
+    end
+
+    it "does not include the summary when masterDefendantId does not match defendantId" do
+      summaries = defendant_summary.application_summaries
+      expect(summaries.first.data["applicationId"]).not_to eq("f38d0030-0b4a-4fa5-9484-bb37b1e6ac56")
```
patrick-laa, line 97: The problem with this test is that if the code didn't work, the test would still pass. You've passed in `...ab39` and `...ac56` to the defendant_summary. If filtering wasn't working, then `defendant_summaries.application_summaries` would contain both `...ab39` and `...ac56`. This test only checks that the _first_ application summary isn't `...ac56`... but if the code was buggy, the first application summary would still almost certainly be `...ab39`, so this would pass!

patrick-laa (approved):

PR #1347, opened 2025-04-25
Stop routing web requests to worker/metrics pods

patrick-laa:
## What
Switch labels on non-web pods so we stop routing HTTP requests to them


alexdesi (approved):
my understanding is that you've made the `matchLabels` explicit,
it sounds a positive semplification ✅

PR #1350, opened 2025-04-29
ACD-666: Start using AWS secrets

patrick-laa:
## What
By pushing secrets into AWS we can unencrypt most of our values files. The exception is prod, because it contains some specific SSL client fingerprints that I believe need to stay secrets, and can't easily be extracted into Secret Manager because they don't get set as env vars

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-666)


helm_deploy/laa-court-data-adaptor/templates/_environment.tpl
```diff
@@ -77,7 +89,10 @@ env:
   - name: AWS_REGION
     value: {{ .Values.sqs_region }}
   - name: SENTRY_DSN
-    value: {{ .Values.sentry_dsn }}
+    valueFrom:
+      secretKeyRef:
+        name: aws-secrets
+        key: sentry_dsn
```
muthusubra, line 95: Are we using the same sentry dsn for all env? Also, I see prod and production Envs in Sentry https://ministryofjustice.sentry.io/projects/laa-court-data-adaptor/?project=5375870
Should we cleanup the envs not used?

patrick-laa: Yes, it looks like CDA is currently using the same DSN for all envs. I haven't changed that, but maybe we should separate it all out.

muthusubra: Happy to leave it as it is. Having one DSN with different environments is probably easier to maintain. But, I see that laa-court-data-api has env specific DSN. May be we can create separate story to group the - laa-court-data-api  DSNs to one.

muthusubra (approved):

PR #1352, opened 2025-04-30
Revert metrics change

patrick-laa:
My last PR broke prometheus metrics collection, because our code expects to be able to make HTTP requests to the metrics pods. I _think_ that because metrics pods only expose a different port to normal pods, normal HTTP requests won't be routed to metrics pods. So I'm reverting the metrics deployment part of my last PR (but keeping the worker deployment bit, as that was I think the most problematic part).


alexdesi (commented):
it looks good ✅

alexdesi (approved):

PR #1353, opened 2025-04-30
ACD-669: Adjust to match HMCTS schema/param differences

patrick-laa:
## What
- Stop expecting court application type to include validTo, resentencingActivationCode and prefix
- Since the hearing results endpoint ignores the sittingDay param, stop sending it and also stop calling it multiple times with different sittingDays
- Also stop loading hearing events when constructing the v1 defendant payload as that slows down the endpoint.

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-669)

## Checklist

### DIFF
--- a/app/models/hearing.rb
+++ b/app/models/hearing.rb
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 
 class Hearing
-  attr_reader :data, :skip_events
+  attr_reader :data, :load_events
 
   delegate :blank?, to: :data
 
-  def initialize(data, skip_events: false)
+  def initialize(data, load_events: true)
     @data = HashWithIndifferentAccess.new(data || {})
-    @skip_events = skip_events
+    @load_events = load_events
   end
 
   def id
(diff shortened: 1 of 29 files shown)


app/models/hearing.rb
```diff
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
 class Hearing
-  attr_reader :data
+  attr_reader :data, :skip_events
 
   delegate :blank?, to: :data
 
-  def initialize(data)
+  def initialize(data, skip_events: false)
```
alexdesi, line 8: is this a double negation? 🤔  **skip** event **false**

What about using direct logic `include_events: true`?
also "include" sounds better

alexdesi (commented):
Nice one!
My one concern relates to the naming of the param `skip_events`,
but If it feels right to you, let's keep it 😄

alexdesi (approved):
`load_events` works well 👍 
✅

PR #1354, opened 2025-05-02
Remove redundant class that was breaking other specs

patrick-laa:
## What
The SchemaValidator class was loading in JSON schemas in a way that was breaking schema validation in other specs (but only if the SchemaValidator specs were run before those other specs). And it turned out that the SchemaValidator class isn't even called anywhere.


alexdesi (approved):
I love this PR 😄 
💚 🧹 💚 🧹 💚 🧹

PR #1356, opened 2025-05-12
ACD-701: Add start date to offences payload

patrick-laa:
## What
Allow us to show offence dates in VCD without having to switch to the v2 defendants endpoint

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-701)


alexdesi (approved):
it looks good ✅ 
I wonder why they used those minimalistic test description  (which does not describe what we are testing)

PR #1360, opened 2025-05-19
ACD-757: Use masterDefendantId to link appeals to defendants

patrick-laa:
## What
- HMCTS defendant summaries actually include masterDefendantId, which is sometimes but not always the same as defendantId
- Since appeals include masterDefendantId but not defendantId, we should consistently compare masterDefendantId to masterDefendantId when working out which defendant is associated with which appeal

[Link to story](https://dsdmoj.atlassian.net/browse/-ACD-757)

### DIFF
--- a/lib/schemas/global/search/apiDefendantSummary.json
+++ b/lib/schemas/global/search/apiDefendantSummary.json
@@ -8,6 +8,10 @@
       "description": "The identifier of the defendant",
       "$ref": "http://justice.gov.uk/core/courts/external/apiCourtsDefinitions.json#/definitions/uuid"
     },
+    "masterDefendantId": {
+      "$ref": "http://justice.gov.uk/core/courts/external/apiCourtsDefinitions.json#/definitions/uuid",
+      "description": "Technical identifier that uniquely identifies the defendant on CPP i.e. across prosecution cases.  Case defendants that are the same person will be matched and provided the same defendantMasterId"
+    },
     "defendantNINO": {
       "description": "The person nino when the defendant is a person",
       "$ref": "http://justice.gov.uk/core/courts/external/apiCourtsDefinitions.json#/definitions/nino"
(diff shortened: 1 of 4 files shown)


lib/schemas/global/search/apiDefendantSummary.json
```diff
@@ -1,88 +1,92 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "http://justice.gov.uk/core/courts/search/external/apiDefendantSummary.json",
+  "description": "A summary of defendant details",
+  "type": "object",
+  "properties": {
+    "defendantId": {
+      "description": "The identifier of the defendant",
+      "$ref": "http://justice.gov.uk/core/courts/external/apiCourtsDefinitions.json#/definitions/uuid"
+    },
+    "masterDefendantId": {
```
patrick-laa, line 11: This file was formatted really wonkily, something to do with character encoding maybe? All I've done is add this masterDefendantId bit.

spec/models/hmcts_common_platform/defendant_summary_spec.rb
```diff
@@ -79,7 +79,7 @@
 
     let(:defendant_summary) do
```
alexdesi, line 80: it looks like this spec test need a context:
```
context "when masterDefendantId matches defendantId"
  ...
  it "includes only summaries"
  ...
  end
end
```

alexdesi (approved):
it looks good ✅ 
minor comment to refactor a test
No obligation.

spec/models/hmcts_common_platform/defendant_summary_spec.rb
patrick-laa: Ah, but in this context masterDefendantId doesn't match defendantId. defendantId isn't read at all.

PR #1362, opened 2025-05-20
ACD-758: Export SQS queues

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-758)

MAAT went offline before CDA did, meaning there are some messages CDA sent to MAAT that were queued up in SQS but haven't been processed (looks like < 100 in each queue). There is a possibility that MAAT will be offline for longer than the 14-day maximum time that these messages will be stored, meaning they will disappear. We therefore need to download these messages so that if they disappear before being processed, we can re-enqueue the data.

This PR adds a rake task that reads the contents of all queues and prints them out as a single blob of JSON to the console. This data can be saved and re-queued in a future rake task.

Note that to ensure that messages aren't duplicated, this task tells SQS to temporarily hide messages it downloads, for a period of 10 seconds. That means that once this task has run there will be an up-to-10-second period when messages aren't visible.


mo-zag (approved):
All good I have approved,  would it be also be an idea to save these messages within a DB table to that we can keep these for longer than 14 days.

lib/tasks/download_queue_data.rake
```diff
@@ -0,0 +1,39 @@
...
+          max_number_of_messages: 10,
+          visibility_timeout: 10,
+        )
+
+        messages.each do |message|
+          ret << message.body
+        end
+
+        break unless messages.any?
+      end
```
alexdesi, line 29: I am aware this is just a "run once" script, feel free to ignore this comment:
A `loop` inside a `.map` makes this difficult to maintain/debug, ideally this should be a function.

PR #1368, opened 2025-05-21
ACD-765: Precommit hooks

patrick-laa:
Check for accidentally committed secrets as a hook that runs on commit and on put to GitHub.


alexdesi (approved):
✅

PR #1370, opened 2025-05-27
ACD-774: Add env vars that status controller uses

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-774)


alexdesi (approved):
nice one! ✅

PR #1371, opened 2025-05-27
ACD-349: Try to reduce n+1 queries

patrick-laa:
## What
Pull all local records out of the DB at once and search through them in-memory to reduce the number of DB calls for defendant- and offence-heavy cases and applications.

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-349)

### DIFF
--- a/app/services/court_application_recorder.rb
+++ b/app/services/court_application_recorder.rb
@@ -9,8 +9,15 @@ class CourtApplicationRecorder < ApplicationService
   def call
     court_application.update!(body: model.data)
 
+    local_records = court_application.court_application_defendant_offences
+                                     .where(application_type: model.application_type,
+                                            defendant_id: model.subject_summary.subject_id)
+                                     .to_a
+
     model.subject_summary.offence_summary.each do |offence|
-      CourtApplicationDefendantOffence.find_or_create_by!(
+      next if local_records.any? { |record| record.offence_id == offence.offence_id }
+
+      CourtApplicationDefendantOffence.create!(
         court_application_id: court_application.id,
         defendant_id: model.subject_summary.subject_id,
         offence_id: offence.offence_id,
--- a/app/services/prosecution_case_recorder.rb
+++ b/app/services/prosecution_case_recorder.rb
@@ -9,9 +9,13 @@ class ProsecutionCaseRecorder < ApplicationService
   def call
     prosecution_case.update!(body:)
 
+    local_records = prosecution_case.prosecution_case_defendant_offences.to_a
+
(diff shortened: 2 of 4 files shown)


app/services/court_application_recorder.rb
```diff
@@ -9,8 +9,15 @@ def initialize(court_application_id, model)
   def call
     court_application.update!(body: model.data)
 
+    local_records = court_application.court_application_defendant_offences
+                                     .where(application_type: model.application_type,
+                                            defendant_id: model.subject_summary.subject_id)
+                                     .to_a
```
alexdesi, line 15: do we need to convert it to array?
(see next comment)

app/services/court_application_recorder.rb
```diff
@@ -9,8 +9,15 @@ def initialize(court_application_id, model)
...
+
     model.subject_summary.offence_summary.each do |offence|
-      CourtApplicationDefendantOffence.find_or_create_by!(
+      next if local_records.any? { |record| record.offence_id == offence.offence_id }
+
```
alexdesi, line 19: if `local_records` was an active_record we could use the `exist?` without a block:

```
unless local_records.exist?(offence_id: offence.offence_id) do
  CourtApplicationDefendantOffence.create!(
```

app/services/prosecution_case_recorder.rb
```diff
@@ -9,9 +9,13 @@ def initialize(prosecution_case_id:, body:)
   def call
     prosecution_case.update!(body:)
 
+    local_records = prosecution_case.prosecution_case_defendant_offences.to_a
+
     prosecution_case.defendants.each do |defendant|
       defendant.offences.each do |offence|
-        ProsecutionCaseDefendantOffence.find_or_create_by!(
+        next if local_records.any? { |record| record.defendant_id == defendant.id && record.offence_id == offence.id }
```
alexdesi, line 16: same comment as before

app/services/court_application_recorder.rb
patrick-laa: The issue we're solving here is an N+1 query, so what we're trying to do is not make a query to the DB in every iteration of the `each` loops. If we leave `local_records` as an ActiveRecord collection, then checking whether a record exists inside it will trigger a DB lookup. As a performance improvement we do `to_a` to pull the records into memory once so we don't need to go back to the DB again.

patrick-laa: Added a comment to explain this.

alexdesi (approved):
it looks good! ✅

PR #1377, opened 2025-05-30
Fix specs

patrick-laa:
## What
PR to get specs passing again. Note I've had to make the schemas regex more specific because the old regex was false positiving on the new MAAT API.


alexdesi (approved):
it looks good! ✅

PR #1378, opened 2025-06-02
ACD-368: Specify maat reference when unlinking

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/-ACD-368)

To make sure we update the correct record in our DB, if the client provides a MAAT reference when unlinking, use that MAAT reference to identify the correct local record to update.


alexdesi (commented):
Nice one! ✅ 

just one thing: `retrieve_laa_reference` is duplicated,
not a big issue, but it'd be nice to "dry" it

alexdesi (approved):
it looks good ✅

PR #1383, opened 2025-06-04
ACD-790: Enable modsec everywhere in detection only mode

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-790)

Switch on Modsec in detection-only mode and tag the logs with our GH team so that in a month we can review and see if ModSec is catching anything in production (unlikely given the new IP whitelisting but you never know), and how much it's false positive-ing.


alexdesi (approved):
It looks good! ✅

PR #1391, opened 2025-06-11
ACD-815: Make explicit when classes are about prosecution cases

patrick-laa:
## What
Standardise the naming of things, so that entities are either `CourtApplication` or `ProsecutionCase`.

Note that this should be deployed to production out-of-hours because we are renaming some async jobs and we don't want to risk enqueueing a job that refers to a class that stops existing when the job is run.

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-815)


alexdesi (approved):
ops I forgot to review this one,
yeah, it sounds good! 🚀

PR #1395, opened 2025-06-12
ACD-821: Switch from async to sync for v2 laa references

patrick-laa:
## What
For v2 LAA reference link/unlinks for prosecution cases, switch from async to sync calls.
Note I've not deleted the worker classes. This is for 2 reasons. First, I've left the v1 endpoint async to minimise the number of unnecessary changes to the code. Second, this means that we don't have to deploy the PR out-of-hours, because at the point of deployment, a caseworker might enqueue an async job and we want the worker to still know how to process it.

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-821)

### DIFF
--- a/postman-collection/runner/tests.postman_collection.json
+++ b/postman-collection/runner/tests.postman_collection.json
@@ -809,8 +809,8 @@
 									"listen": "test",
 									"script": {
 										"exec": [
-											"pm.test(\"Status code is 422 or 202\", function () {",
-											"    pm.expect(pm.response.code).to.be.oneOf([422, 202]);",
+											"pm.test(\"Status code is 422 or 201\", function () {",
+											"    pm.expect(pm.response.code).to.be.oneOf([422, 201]);",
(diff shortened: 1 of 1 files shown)


app/controllers/api/internal/v2/laa_references_controller.rb
```diff
@@ -9,22 +9,21 @@ def create
           contract = NewLaaReferenceContract.new.call(**transformed_params)
           enforce_contract!(contract)
 
-          MaatLinkCreatorWorker.perform_async(
-            Current.request_id,
+          MaatLinkCreator.call(
             transformed_params[:defendant_id],
             transformed_params[:user_name],
             transformed_params[:maat_reference],
           )
```
alexdesi, line 16: Should we check for the result?
that return errors, if I am not wrong as exceptions,
maybe we should rescue the exception and based on the code return different error message code?

PR #1396, opened 2025-06-13
transform Link/Unlink for appellants calls, from async to sync calls

alexdesi:
## What
convert  the link/unlink, from Async to Sync operations

### Ticket
https://dsdmoj.atlassian.net/browse/ACD-813


spec/requests/api/internal/v2/court_application_laa_references_spec.rb
```diff
@@ -39,7 +31,7 @@
       security [{ oAuth: [] }]
       parameter "$ref" => "#/components/parameters/transaction_id_header"
 
-      response(202, "Accepted") do
+      response(201, "Created") do
```
patrick-laa, line 34: This file is used to generate the swagger yaml, so you'll want to run `rails rswag` to update it.

app/controllers/api/internal/v2/court_application_laa_references_controller.rb
```diff
@@ -6,16 +6,28 @@ def create
...
-          head :accepted
+          CourtApplicationLaaReferenceUnlinker.call(
+            subject_id: transformed_params[:subject_id],
+            user_name: transformed_params[:user_name],
+            unlink_reason_code: transformed_params[:unlink_reason_code],
+            unlink_other_reason_text: transformed_params[:unlink_other_reason_text],
+            maat_reference: transformed_params[:maat_reference],
+          )
+
+          head :ok
```
patrick-laa, line 30: Mo's postman tests are still expecting 202s here - you'll need to update them to get them to pass.

spec/requests/api/internal/v2/court_application_laa_references_spec.rb
```diff
@@ -161,35 +166,38 @@
...
-            let(:subject_id) { "fa7ca7bd-5dce-419c-88db-f42e1b7ce8a0" }
+          let(:Authorization) { "Bearer #{token.token}" }
+          let(:subject_id) { "fa7ca7bd-5dce-419c-88db-f42e1b7ce8a0" }
 
+          response("404", "Defendant not found") do
             run_test! do |response|
               error = JSON.parse(response.body)["error"]
 
-              expect(error).to include("Couldn't find LaaReference")
+              expect(error).to include("Defendant not found!")
```
patrick-laa, line 200: When the call to Common Platform fails, what status code do you want to be returned? You should add a spec demonstrating that to this file. (Hint, for Prosecution Case LAA References we return a 424, and you can see the spec at https://github.com/ministryofjustice/laa-court-data-adaptor/blob/main/spec/requests/api/internal/v2/laa_references_spec.rb#L65

alexdesi: I've fixed it this way: any possible connection error with CP, will be converted to 424 - Failed Dependency.

patrick-laa: I'd be wary of trying to list every possible Faraday error in the ApplicationController. For one thing as currently configured, if CP returns a 404 status code that _won't_ raise a Faraday::ResourceNotFound, and the service layer will raise an unhandled StandardError. For another, Faraday errors don't have `error_code` attributes, so won't be handled very nicely, whereas our own CommonPlatform::Errors::FailedDependency _does_ have an error code. So ideally you'd catch any errors lower down and wrap them up into our own error class.

alexdesi: ok, it makes sense,  I'll chage it to raise the`CommonPlatform::Api::Errors::FailedDependency` 👍

app/controllers/api/internal/v2/court_application_laa_references_controller.rb
alexdesi: Update: I had to include the 404 status as well, since it is a sync, the not found is also a possible status now.

PR #1402, opened 2025-06-19
ACD-834: Save unlink_reason_code

patrick-laa:
## What
Start logging the unlink reason code as this will allow us to analyse better why users are unlinking
[Link to story](https://dsdmoj.atlassian.net/browse/ACD-834)


app/models/laa_reference.rb
```diff
@@ -5,8 +5,8 @@ class LaaReference < ApplicationRecord
   validates :maat_reference, presence: true, uniqueness: { conditions: -> { where(linked: true) } }
   validates :user_name, presence: true
```
alexdesi, line 7: It may be worth to add validation for `unlink_reason_code`, for example: 
`validates :unlink_reason_code, numericality: { only_integer: true, greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }`

alexdesi (approved):
It looks good,
but probably we should add validation for the `unlink_reason_code` (to avoid misuse of the api)

PR #1403, opened 2025-06-23
ACD-621 list defendants with multiple MAATs

patrick-laa:
## What
Our `prosecution_cases` table contains copies of HMCTS data. Some of this data is corrupt in that different offences for the same defendant will have different MAAT references. HMCTS have requested we give them a list of every case we know about where this happens, along with the correct value. Early tests indicate ~70k affected records, and it's hard to detect the issue without doing analysis on a JSON blob. So this PR tries to find a sane way of running the query as follows:
- A rake task kicks off an async job that can run on a worker pod
- This job creates some temporary views that represent the dodgy data in table form
- It then writes the data to a text file
- Then cleans up after itself

We can then retrieve the text file via a `kubectl cp` command.


app/workers/multiple_maat_analysis_worker.rb
```diff
@@ -0,0 +1,28 @@
...
+    File.open(Rails.root.join("tmp/records_with_multiple_maats.txt"), "w") do |file|
+      file.write(data.first.keys.join("|"))
+      file.write "\n"
+      data.each do |row_with_issue|
+        file.write(row_with_issue.values.join("|"))
+        file.write "\n"
+      end
+    end
+
+    execute_script :delete_views
```
alexdesi, line 17: What about commenting this out?
it could be useful to debug potential issues

alexdesi (approved):
it looks good, 
but of course those complex SQL must be testes,  
especially the `create_views`.

I'd comment out the delete_view for now

✅

PR #1406, opened 2025-06-24
ACD-843: Handle it when CP returns 2 results for a URN

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-843)
 Our v1 defendant finder code wrongly assumes that a search for a specific URN will return either 0 or 1 results. But because searches are case-insensitive, and CP has cases whose URNs differ _only_ in their case, this assumption is sometimes wrong.
 
 This PR fixes that and, when CP returns multiple results, prefers the one that matches the given URN exactly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,31 @@ updates:
   open-pull-requests-limit: 15
   allow:
     - dependency-type: "all"
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: weekly
+    day: tuesday
+    time: "03:00"
+    timezone: Europe/London
+  groups:
+    actions:
+      patterns:
+        - "*"
+  open-pull-requests-limit: 15
+  allow:
+    - dependency-type: "all"
+- package-ecosystem: docker
+  directory: "/"
+  schedule:
+    interval: weekly
+    day: monday
+    time: "03:00"
+    timezone: Europe/London
+  groups:
+    docker:
+      patterns:
+        - "*"
+  open-pull-requests-limit: 15
+  allow:
+    - dependency-type: "all"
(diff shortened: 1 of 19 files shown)


app/services/common_platform/api/defendant_finder.rb
```diff
@@ -76,6 +76,15 @@ def detect_spaces_error
                                            :spaces_in_urn)
       end
 
+      def load_from_common_platform(urn)
+        results = CommonPlatform::Api::SearchProsecutionCase.call(prosecution_case_reference: urn)
+
+        # Very occasionally this case-insensitive URN search will return 2 results, and in that
+        # case we prefer the one that matches the URN exactly. Failing that we just take the first
+        # result
+        results.find { _1.body["prosecutionCaseReference"] == urn } || results.first
```
alexdesi, line 85: Ruby 3.4 introduce `it` as block param, so we can use this now:
`results.find { it.body["prosecutionCaseReference"] }`
which is more readable.

https://www.mintbit.com/blog/writing-more-expressive-ruby-with-the-it-shorthand/

alexdesi: it's not clear (at lest to me 😄  ) why
`results.find { it.body["prosecutionCaseReference"] == urn } ` could possibly be `nil ` ?

patrick-laa: Oh yes, "it" is great! I will change it.

The issue I was worrying about is if `urn` is "TEST12345" but the only result has prosecutionCaseReference "test12345". My worry was that in some cases these are treated as separate URNs but in some they are the same. BUT thinking about it, since the `urn` we use comes from our local copy of Common Platform data, it should _always_ match a Common Platform record, so you're right, it should never be nil.  I've updated the comment and removed the fallback.

alexdesi (approved):
Perfect! 🚀 ✅

PR #1410, opened 2025-06-25
ACD-841: Stop filtering prosecution concluded payloads

patrick-laa:
## What
Common Platform is sending us a bunch of data that MAAT needs but that we don't know about so we're filtering out. Crime Apps have requested that we send them everything, so this PR removes the filtering

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-841)


alexdesi (approved):
it looks good ✅

PR #1411, opened 2025-06-26
ACD-842: Basic batch repull API

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-842)

Accept an arbitrary number of MAAT IDs, work out which cases they relate to, and asynchronously repull all hearings related to those cases, with an endpoint to show the progress of the repulls to


alexdesi (approved):
it looks good ✅

PR #1414, opened 2025-07-01
Try to capture Sidekiq errors in Sentry

patrick-laa:
We're not seeing Sentry notifications when Sidekiq jobs fail, either temporarily or permanently. This isn't great. So this PR attempts to make Sentry be told every time a job fails, even transient errors (like Sentry is notified of transient errors for web requests).


alexdesi (approved):
it looks good! ✅

PR #1416, opened 2025-07-03
ACD-857: Standardise Sentry config

patrick-laa:
## What
- Remove `sample_rate` setting which was causing most CDA errors not to be reported to Sentry 95% of the time 😱 

Then also
- Explicitly tell Sentry which environment we're talking about
- Reduce the amount of performance profiling traffic we sent to Sentry by skipping the status controller endpoints


alexdesi (approved):
nice catch! 🚀 ✅

PR #1418, opened 2025-07-03
ACD-863: Stop silently failing

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-863)

If VCD passes an invalid MAAT ID to the unlink endpoint, we currently silently fail. Instead we should return an error, otherwise VCD tells the user that unlinking has succeeded. This PR makes `LaaReference. retrieve_by_defendant_id_and_optional_maat_reference` raise errors if an invalid MAAT ID is passed, and removes the early returns from the unlinker services

### DIFF
--- a/app/models/laa_reference.rb
+++ b/app/models/laa_reference.rb
@@ -39,13 +39,8 @@ class LaaReference < ApplicationRecord
 
   def self.retrieve_by_defendant_id_and_optional_maat_reference(defendant_id, maat_reference = nil)
     laa_refs = where(defendant_id:, linked: true)
+    ref = maat_reference.blank? ? laa_refs.first : laa_refs.find_by(maat_reference:)
 
-    if laa_refs.blank?
-      raise ActiveRecord::RecordNotFound, "Defendant not found or already unlinked!"
-    end
-
-    return laa_refs.first if maat_reference.blank?
-
-    laa_refs.find_by(maat_reference:)
+    ref || (raise ActiveRecord::RecordNotFound, "Defendant not found or already unlinked!")
   end
 end
--- a/spec/models/laa_reference_spec.rb
+++ b/spec/models/laa_reference_spec.rb
@@ -109,4 +109,42 @@ RSpec.describe LaaReference, type: :model do
       expect(described_class.new(maat_reference: nil)).not_to be_dummy_maat_reference
     end
   end
+
+  describe ".retrieve_by_defendant_id_and_optional_maat_reference" do
+    before { laa_reference.save! }
+
+    context "when only defendant id provided" do
+      subject(:retrieve) { described_class.retrieve_by_defendant_id_and_optional_maat_reference(defendant_id) }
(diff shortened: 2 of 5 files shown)


spec/models/laa_reference_spec.rb
```diff
@@ -109,4 +109,42 @@
       expect(described_class.new(maat_reference: nil)).not_to be_dummy_maat_reference
     end
   end
+
+  describe ".retrieve_by_defendant_id_and_optional_maat_reference" do
+    before { laa_reference.save! }
+
+    context "when only defendant id provided" do
+      subject(:retrieve) { described_class.retrieve_by_defendant_id_and_optional_maat_reference(defendant_id) }
```
alexdesi, line 117: instead of `retrieve`, should we name it accordinly to the retrieved subject?
maybe `defendant` ?

app/models/laa_reference.rb
```diff
@@ -39,13 +39,8 @@ def self.generate_unlinking_dummy_maat_reference
...
+    ref = maat_reference.blank? ? laa_refs.first : laa_refs.find_by(maat_reference:)
 
-    if laa_refs.blank?
-      raise ActiveRecord::RecordNotFound, "Defendant not found or already unlinked!"
-    end
-
-    return laa_refs.first if maat_reference.blank?
-
-    laa_refs.find_by(maat_reference:)
+    ref || (raise ActiveRecord::RecordNotFound, "Defendant not found or already unlinked!")
```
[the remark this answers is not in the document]
patrick-laa: Good idea! I've had to refactor so that we use a bang both if there is a maat_reference and if there isn't, and then also to make sure if there is a MAAT reference we make clear that that could be the thing that stopped us finding the record.

spec/models/laa_reference_spec.rb
patrick-laa: So it's an LAA reference, not a defendant, that we're retrieving. And there's already `let` called `laa_reference` in this spec, so I've gone with `retrieved_laa_reference`

app/models/laa_reference.rb
```diff
@@ -38,14 +38,11 @@ def self.generate_unlinking_dummy_maat_reference
...
-
-    return laa_refs.first if maat_reference.blank?
-
-    laa_refs.find_by(maat_reference:)
+    filters = { defendant_id:, linked: true }
+    filters[:maat_reference] = maat_reference if maat_reference.present?
+    find_by!(filters)
+  rescue ActiveRecord::RecordNotFound
+    attributes = filters.except(:linked).map { |k, v| "#{k}: '#{v}'" }.join(" and ")
+    raise ActiveRecord::RecordNotFound, "LAA reference with #{attributes} not found or already unlinked!"
```
alexdesi, line 46: nice and clean! ✨

alexdesi (approved):
it looks good! ✅

PR #1419, opened 2025-07-04
Fix/refactor Unlink defendant

alexdesi:
## What
Refactor Unlink defendant, alligning the code to other controller/actions/
As a side effect - which is correct - return code for invalid input is 422, and not 400.

## Jira
https://dsdmoj.atlassian.net/browse/ACD-852

### DIFF
--- a/app/contracts/prosecution_case_laa_reference_contract.rb
+++ b/app/contracts/prosecution_case_laa_reference_contract.rb
@@ -13,7 +13,9 @@ class ProsecutionCaseLaaReferenceContract < Dry::Validation::Contract
 
   rule(:defendant_id) do
     key.failure("is not a valid uuid") unless uuid_validator.call(uuid: value)
-    key.failure("cannot be linked right now as we do not have all the required information, please try again later") unless link_validator.call(defendant_id: value)
+    unless link_validator.call(defendant_id: value)
+      key.failure("cannot be linked right now as we do not have all the required information, please try again later")
+    end
   end
 
   rule(:maat_reference) do
--- a/app/controllers/api/internal/v2/prosecution_case_laa_references_controller.rb
+++ b/app/controllers/api/internal/v2/prosecution_case_laa_references_controller.rb
@@ -4,7 +4,7 @@ module Api
   module Internal
     module V2
       class ProsecutionCaseLaaReferencesController < ApplicationController
-        # Create link to court data
+        # Link the Defendant to court data
         def create
           contract = ProsecutionCaseLaaReferenceContract.new.call(**transformed_params)
           enforce_contract!(contract)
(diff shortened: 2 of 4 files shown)


app/contracts/prosecution_case_laa_reference_contract.rb
```diff
@@ -13,7 +13,9 @@ class ProsecutionCaseLaaReferenceContract < Dry::Validation::Contract
 
   rule(:defendant_id) do
     key.failure("is not a valid uuid") unless uuid_validator.call(uuid: value)
-    key.failure("cannot be linked right now as we do not have all the required information, please try again later") unless link_validator.call(defendant_id: value)
+    unless link_validator.call(defendant_id: value)
+      key.failure("cannot be linked right now as we do not have all the required information, please try again later")
+    end
```
alexdesi, line 18: - Line too long
- Improve readability

app/controllers/api/internal/v2/prosecution_case_laa_references_controller.rb
```diff
@@ -37,18 +35,12 @@ def enforce_contract!(contract)
...
-        end
-
-        def check_defendant_presence!
-          laa_reference = LaaReference.find_by(defendant_id: transformed_params[:defendant_id])
-
-          if laa_reference.nil?
-            raise ActiveRecord::RecordNotFound, "Defendant not found!"
-          end
+        rescue ActiveRecord::RecordNotFound
+          raise ActiveRecord::RecordNotFound, "Defendant not found!"
```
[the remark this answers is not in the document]
alexdesi: I am not totally sure we are doing that 😄 
I've just refactored it but I left the same functionality.
I'd be happy to remove the rescue, ActiveRecord will take care of building the error message with the proper params.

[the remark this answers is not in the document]
patrick-laa: I guess technically there are 2 ways this could fail - VCD could provide an invalid defendant ID, so 'Defendant not found!' is appropriate. But if for some reason there is no LAA reference, e.g. the defendant was never linked in the first place, then 'LAA reference not found' is more appropriate. But it doesn't hugely matter either way, as it's an internal error message.

app/controllers/api/internal/v2/prosecution_case_laa_references_controller.rb
```diff
@@ -37,18 +35,13 @@ def enforce_contract!(contract)
           end
         end
 
-        def unlink
-          check_defendant_presence!
+        def unlink_defendant!
+          defendant_id = transformed_params[:defendant_id]
```
patrick-laa, line 39: Thinking about it, you could simply call `LaaReference.retrieve_by_defendant_id_and_optional_maat_reference(transformed_params[:defendant_id]` here, as that will now raise an `ActiveRecord::RecordNotFound` with an appropriate error message all by itself.

alexdesi: good idea, something for later! ... I'm releasing it now :)

PR #1422, opened 2025-07-08
ACD-854: Remove unused endpoints

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-854)

Remove all unused internal v1 endpoints (i.e. all except defendants and rep orders)
Remove all external v2 endpoints


.env.test
```diff
@@ -9,7 +9,7 @@ AWS_UNLINK_QUEUE_URL=
 AWS_HEARING_RESULTED_QUEUE_URL=
 AWS_PROSECUTION_CONCLUDED_QUEUE_URL=
 
-INLINE_SIDEKIQ=true
+INLINE_SIDEKIQ=false
```
patrick-laa, line 12: There was a global `Sidekiq::Testing.fake!` hidden in one of the now-deleted spec files that caused sidekiq to _not_ be inline for the entire test suite. To avoid breaking all the tests, I've updated this to have the same effect as the now-deleted config call.

app/controllers/api/external/v1/prosecution_conclusions_controller.rb
```diff
@@ -2,7 +2,84 @@ module Api
   module External
     module V1
       class ProsecutionConclusionsController < ApplicationController
-        include ProsecutionConcludable
```
patrick-laa, line 5: Now that this is only used once, no point having the logic in a concern

config/routes.rb
```diff
@@ -19,12 +19,9 @@
   namespace :api do
     namespace :internal do
       api_version(module: "V1", path: { value: "v1" }, default: true) do
-        resources :prosecution_cases, only: [:index]
-        resources :prosecution_case_laa_references, path: "laa_references", only: %i[create]
         resources :defendants, only: %i[update show]
         resources :prosecution_case_representation_orders, path: "representation_orders", only: [:create]
         resources :court_application_representation_orders, only: [:create]
```
alexdesi, line 24: I've removed `court_application_representation_orders` as well, in a different PR, anyway this was open before than mine.
I believe you can also remove the fixture file `spec/fixtures/files/court_application_search_result.json`

alexdesi (commented):
I think it's ok, but you should remove the fixture too,
for example`spec/fixtures/files/court_application_search_result.json`
and potentially more fixtures.
have a look to all the `file_fixture` to identify the files.

config/routes.rb
patrick-laa: I've been through every deleted `file_fixture` call to check which fixtures are no longer referenced.

alexdesi (approved):
I love this! 🧹 ✨

PR #1424, opened 2025-07-09
ACD-872: Allow defendant page to load even if URN has a space.

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-872)

Previously, if a caseworker searched by defendant details and Common Platform returned a record with a space in the URN, when the caseworker clicked through to the defendant details page we would try to look up the record by URN and this fails because Common Platform strips the spaces from search terms so doesn't match URNs that do have spaces.

This PR adds a fallback so that if we can't load the record by URN we will fall back to look up by defendant details, using the defendant details in our local DB. At every stage it checks to make sure that what is returned actually matches what was searched for.

### DIFF
--- a/app/controllers/api/internal/v2/defendants_controller.rb
+++ b/app/controllers/api/internal/v2/defendants_controller.rb
@@ -5,13 +5,14 @@ module Api
     module V2
       class DefendantsController < ApplicationController
         def show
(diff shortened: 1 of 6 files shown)


alexdesi (approved):
it looks good, just minor comments

PR #1426, opened 2025-07-10
Handle unrecognised court centre IDs gracefully

patrick-laa:
## What
When no explicit code is returned, we try to look up the code based on the ID from a hard-coded list in a gem. But if that fails we currently throw an error even though other parts of the code make clear that we should be able to cope with not knowing what the code is.


alexdesi (approved):
Nice and clear! 🚀 ✅

PR #1428, opened 2025-07-11
Send appropriate messages to MAAT

patrick-laa:
## What
When a court application hearing is resulted, Common Platform was already sending us messages, and we were already sending messages about court applications associated with the hearing to MAAT. So we thought no action was needed. BUT it turns out we were only sending messages about MAAT-linked defendants of the case associated with the appeal, not the subject of the appeal. This PR fixes that.


alexdesi (approved):
it looks good, I've added a message to understand in which case `laa_reference` is nil and why?

PR #1429, opened 2025-07-11
Remove court_application_representation_orders endpoint.

alexdesi:
## What
Remove court_application_representation_orders endpoint.

Jira: https://dsdmoj.atlassian.net/browse/ACD-874


alexdesi:
I guess we can close this, since this is included in another PR.

PR #1431, opened 2025-07-15
ACD-885: Strip out the unicode characters SQS cannot handle

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-885)


alexdesi (approved):
Nice one!
if you can, amend the message (SQS does not striclty fails).

PR #1436, opened 2025-07-21
ACD-892: Clear out redundant PCDOs

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-892)

Automatically get rid of old ProsecutionCaseDefendantOffences that no longer correspond to what Common Platform tells us about defendants and offences,


spec/services/prosecution_case_recorder_spec.rb
```diff
@@ -61,5 +61,46 @@
...
+
+    context "when other prosecution case exists" do
+      let!(:other_prosecution_case) do
+        ProsecutionCase.create!(
+          id: SecureRandom.uuid,
+          body:,
+        )
+      end
+
+      context "when this case has a PCDO for the other case's defendant" do
```
alexdesi, line 73: First time you use PCDO it's good to be explicit, then you can use the acronimo, for example:
"when this case has a PCDO (Prosecution Case Defendant Offence) for the other case's defendant"

alexdesi (approved):
Nice one! 
just minor comments

PR #1441, opened 2025-07-23
ACD-901: Stop squashing court applications into prosecution case structures

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-901)

Resolve the following tech debt items:

* ProsecutionCaseDefendantOffences is a messy mechanism - it's hard to keep them in date with CommonPlatform. So since we don't need to use them for Court Applications, we can stop using them
* Court Applications shouldn't live in a table called `prosecution_cases` which already contains prosecution cases
* We shouldn't pretend that court applications have many defendants when in fact they have one subject. It's better to store the subject ID directly against the court application object.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -92,7 +92,7 @@ GEM
       public_suffix (>= 2.0.2, < 7.0)
     ast (2.4.3)
     aws-eventstream (1.4.0)
-    aws-partitions (1.1133.0)
+    aws-partitions (1.1137.0)
     aws-sdk-core (3.227.0)
       aws-eventstream (~> 1, >= 1.3.0)
       aws-partitions (~> 1, >= 1.992.0)
@@ -100,7 +100,7 @@ GEM
       base64
       jmespath (~> 1, >= 1.6.1)
       logger
-    aws-sdk-sqs (1.97.0)
+    aws-sdk-sqs (1.98.0)
       aws-sdk-core (~> 3, >= 3.227.0)
       aws-sigv4 (~> 1.5)
     aws-sigv4 (1.12.1)
@@ -171,7 +171,7 @@ GEM
     erb (5.0.2)
     erubi (1.13.1)
     erubis (2.7.0)
-    faraday (2.13.3)
+    faraday (2.13.4)
       faraday-net_http (>= 2.0, < 3.5)
       json
       logger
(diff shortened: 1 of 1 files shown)


app/services/common_platform/api/representation_order_creator.rb
```diff
@@ -18,45 +18,34 @@ def call
...
-              status_date:,
-              effective_start_date:,
-              effective_end_date:,
-              defence_organisation:,
+              **offence.merge(
+                court_application_id: court_application.id,
+                subject_id: defendant_id,
+                application_reference: maat_reference,
+                defence_organisation:,
+              ),
```
alexdesi, line 29: After 10 min, I realised:
`**` return the unpacked result from merge and pass it into `RecordCourtApplicationRepresentationOrder.call(`  as `keyword arguments` 

That's smart, but I try to be more in the maintainable/boring/explicit side:
Maybe something like this?

```
offence.merge(
                case_defendant_offence:,
                defendant_id:,
                defence_organisation:,
                application_reference: maat_reference,
              )

CommonPlatform::Api::RecordProsecutionCaseRepresentationOrder.call(**offence)

```

app/services/common_platform/api/representation_order_creator.rb
```diff
@@ -18,45 +18,34 @@ def call
...
-              case_defendant_offence:,
-              defendant_id:,
-              offence_id:,
-              status_code:,
-              application_reference:,
-              status_date:,
-              effective_start_date:,
-              effective_end_date:,
-              defence_organisation:,
+              **offence.merge(
```
alexdesi, line 35: same here

app/services/common_platform/api/representation_order_creator.rb
```diff
@@ -18,45 +18,34 @@ def call
...
+                case_defendant_offence:,
+                defendant_id:,
+                defence_organisation:,
+                application_reference: maat_reference,
+              ),
```
alexdesi, line 40: Anyway, do you need the comma here?

spec/services/court_application_link_validator_spec.rb
```diff
@@ -6,35 +6,21 @@
...
-    end
-
-    context "when the hearing summary does not exist" do
-      before do
-        court_application.body.delete("hearingSummary")
-        court_application.save!
-      end
+  it "returns true" do
+    expect(link_validator_response).to be true
+  end
```
alexdesi, line 16: Nit:
`link_validator_response` is the subject, you could do:

`it { is_expected.to be true }`

alexdesi (approved):
This is great!  I love this simplification, but it looks like a considerable change.
Be sure to test extensively, 
potentially we could start using the E2E tests.

I added comments about making the code more explicit.

PR #1443, opened 2025-07-24
ACD-361: Fix fingerprinting config

patrick-laa:
## What
It looks to me like the SSL fingerprint verification config we had set up in the external ingress for prod was misconfigured. I have been testing the config in the `test` env and am now moving across to prod what seems to work

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-361)


alexdesi (approved):
It sounds great to me.

PR #1457, opened 2025-08-26
ACD-912: (Appeals v2) New endpoint for "Related court applications" tab

patrick-laa:
## What
[Link to story](https://dsdmoj.atlassian.net/browse/ACD-912)


app/models/hmcts_common_platform/court_application_summary.rb
```diff
@@ -83,6 +87,7 @@ def to_builder
         summary.hearing_summary hearing_summary.map(&:to_json)
         summary.subject_summary subject_summary.to_json
         summary.judicial_results judicial_results.map(&:to_json)
+        summary.maat_id maat_id
```
alexdesi, line 90: I wonder if the fact that `maat_id` might be nil could be confusing,
More precisely, `maat_id` might be present but not linked, and that would set maat_id to nil 
Should we call the attribute linked_maat_id ?
Or maybe add a linked attribute?
🤔

app/models/prosecution_case.rb
```diff
@@ -79,4 +84,11 @@ def hearing_results(defendant_id = nil, load_events: true)
   def hearing_summaries_for(defendant_id)
     hearing_summaries.select { |summary| defendant_id.nil? || summary.defendant_ids.include?(defendant_id) }
   end
+
+  def retrieve_full_court_application(application_summary)
+    response = CommonPlatform::Api::CourtApplicationSearcher.call(
+      application_id: application_summary["applicationId"],
+    )
```
alexdesi, line 91: Should we gracefully fail here?
returning nil in case of 4xx or 5xx 🤔

PR #1472, opened 2025-09-18
ACD-988: Remove jsonapi_parameters gem

patrick-laa:
## What
Since we only use it for one endpoint I decided just to handle jsonapi style params manually rather than use a replacement library.

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-988)


app/controllers/api/internal/v1/representation_orders_controller.rb
```diff
@@ -35,7 +35,9 @@ def create_params
               effective_start_date
               effective_end_date
             ],
-          )
+          ).tap do |filtered|
```
alexdesi, line 38: nit: a bit more explicit for the param's name:
` ).tap do |filtered_params|`
or
` ).tap do |permitted_params|`

config/initializers/mime_types.rb
```diff
@@ -0,0 +1,11 @@
...
+
+# In order to support the JSONAPI standard for v1 endpoints we need
+# to register this custom mime type
+JSONAPI_MIME_TYPES = %w[
+  application/vnd.api+json
+  text/x-json
+  application/json
+].freeze
+
+Mime::Type.register JSONAPI_MIME_TYPES.first, :json, JSONAPI_MIME_TYPES
```
alexdesi, line 11: 👍 I had no idea about this

alexdesi (approved):
It looks good,
it's nice we don't need to use another gem!

PR #1474, opened 2025-09-24
ACD-996: Store incoming payloads

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/ACD-996)

### DIFF
--- a/app/services/incoming_payload_logger.rb
+++ b/app/services/incoming_payload_logger.rb
@@ -1,18 +1,19 @@
 class IncomingPayloadLogger < ApplicationService
-  def initialize(body, request_id, payload_type)
+  def initialize(body, request_id, payload_type, identifier)
     @body = body
     @request_id = request_id
     @payload_type = payload_type
+    @identifier = identifier
   end
 
-  attr_reader :body, :request_id, :payload_type
+  attr_reader :body, :request_id, :payload_type, :identifier
 
   def call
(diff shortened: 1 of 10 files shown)


app/services/incoming_payload_logger.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  attr_reader :body, :request_id, :payload_type
+
+  def call
+    IncomingPayload.create!(
+      payload_type:,
+      body:,
+      request_id:,
+      compressed_body: Base64.encode64(Zlib::Deflate.deflate(body.to_json)),
```
patrick-laa, line 15: In real life we'd either want to store `body` _or_ `compressed_body` depending on whether we want to optimise for ease of querying or minimising database size

alexdesi: So, I think we do NOT need explicit compression:
once we deflate + Base64, Postgres just sees a string of gibberish and we **cannot** query body->>'ip' or use JSON operators.

In support of not doing explicit compression:
1. I've just discovered that Postgress alreay apply a compression to `jsonb` for body > 2kb, it's called TOAST: https://www.crunchydata.com/blog/postgres-toast-the-greatest-thing-since-sliced-bread
2. In the future we can easily prune the table, for example remove entries older than 2 years, this way the tables/indecs will be under control.

PR #1520, opened 2026-04-16
ACD-1375: add GET endpoint for XHIBIT migrated cases

alexdesi:
## Summary

Exposes a new internal API endpoint that allows LAA applications to retrieve XHIBIT migrated cases, with support for status filtering and pagination. 
This enables VCD to query and display migrated cases.

- Added `GET /api/internal/v2/link_migrated_cases` endpoint with optional `status` query parameter (`pending`, `auto_linked`, `manually_linked`, `action_required`) and `page`/`per_page` pagination (default 10, max 100)
- Added `status` enum to `XhibitMigratedCase` model
- Added request spec covering default pagination, status filtering, custom pagination, boundary values, and 401 unauthorised

### DIFF
--- a/app/models/xhibit_migrated_case.rb
+++ b/app/models/xhibit_migrated_case.rb
@@ -1,5 +1,11 @@
 class XhibitMigratedCase < ApplicationRecord
   attribute :status, :string, default: "pending"
+  enum :status, {
+    pending: "pending",
+    auto_linked: "auto_linked",
+    manually_linked: "manually_linked",
+    action_required: "action_required",
+  }
 
   validates :case_urn, presence: true
   validates :xhibit_case_number, presence: true
--- a/app/models/xhibit_migrated_case.rb
+++ b/app/models/xhibit_migrated_case.rb
@@ -7,7 +7,12 @@ class XhibitMigratedCase < ApplicationRecord
     action_required: "action_required",
   }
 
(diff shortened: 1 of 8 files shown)


app/models/xhibit_migrated_case.rb
```diff
@@ -1,5 +1,11 @@
 class XhibitMigratedCase < ApplicationRecord
   attribute :status, :string, default: "pending"
+  enum :status, {
+    pending: "pending",
+    auto_linked: "auto_linked",
+    manually_linked: "manually_linked",
+    action_required: "action_required",
+  }
```
sophiebrown01, line 8: This isn't perhaps the place to ask as these are the API changes, but will we be showing the status in a more user-friendly way on VCD (for example, _Action required_ etc)?

alexdesi: Yes, I agree on the more user-friendly way, we can do that in the FE/VCD.
This is helpful in the model for 2 reasons:
- it enforces the status: 
  `case.status = "initial"` would trigger an error (for example0
-  we can use many helpers in the logic, like `case.pending?`, `case.auto_linked?` 

PS. Traditionally, enums used integers, but it's safer to use string (no accidental remapping issues)

sophiebrown01 (approved):
Very comprehensive, great job Alessandro! 🚀

PR #1521, opened 2026-04-20
set ips.restrict to false in UAT

alexdesi:
Disables IPS IP restriction in the UAT environment by setting `ips.restrict: false` in the UAT Helm values.

This allows UAT traffic to bypass the IP allowlist, enabling broader access for testing purposes.


sophiebrown01 (approved):
Looks good (minus the PR title 🙂)

PR #1522, opened 2026-04-21
Add new HMCTS IP addresses to UAT allowlist

alexdesi:
## Summary

Adds two new HMCTS IP addresses (`51.140.138.86` and `85.210.17.233`) to the IP allowlist in the UAT environment, ensuring traffic from these endpoints is permitted through the IP restriction layer.


sophiebrown01 (approved):
Looks good 👍

PR #1523, opened 2026-04-27
Repoint UAT to SIT environment

sophiebrown01:
This PR repoints the UAT environment for CDA to the Common Platform SIT environment.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.4.4-alpine3.20
+FROM ruby:3.4.9-alpine3.23
```
alexdesi, line 1: great,
is this the only place?

alexdesi (approved):
looks good to me! ✅

Dockerfile
sophiebrown01: Just updated the .ruby-version file too 🙂

PR #1524, opened 2026-04-27
Remove hmcts_common_platform gem and inline CSV lookup

alexdesi:
The `hmcts_common_platform` gem is no longer maintained and has been removed.

The gem provided a CSV-based fallback lookup for court centre OU codes (`oucode`), used when the HMCTS Common Platform API omits the `code` field from a court centre payload. 

The gem's functionality has been inlined into the app:
- `lib/organisation_unit.csv` copied from the gem (513 court centre records)
- `spec/models/hmcts_common_platform/reference/court_centre_spec.rb` specs for the inlined class


Gemfile
```diff
@@ -23,7 +23,6 @@ gem "dry-validation"
 gem "faraday", "~> 2.13"
 gem "faraday-net_http_persistent", "~> 2"
 gem "faraday-retry", "~> 2.3"
-gem "hmcts_common_platform", github: "ministryofjustice/hmcts_common_platform", tag: "v0.2.0"
```
alexdesi, line 26: this is the reason of this PR,
removing the unmantained gem hmcts_common_platform

app/models/hmcts_common_platform/reference/court_centre.rb
```diff
@@ -0,0 +1,29 @@
...
+        @row = row
+      end
+
+      # Organisational Unit Code
+      def oucode
+        @row["oucode"]
+      end
+    end
+  end
+end
```
alexdesi: In short, this small class replaced the only functionality of the entire Gem that was still used.

app/models/hmcts_common_platform/reference/court_centre.rb
```diff
@@ -0,0 +1,29 @@
...
+      def self.csv
+        @csv ||= CSV.read(Rails.root.join("lib/organisation_unit.csv"), headers: true)
+      end
+
+      def initialize(row)
+        @row = row
+      end
+
+      # Organisational Unit Code
+      def oucode
```
sophiebrown01: Does this need to be named the same as it was in the gem, `short_oucode`?

sophiebrown01: Additionally you might need to limit what's returned as the previous code did

```
def short_oucode
  oucode[0..4]
end
```

sophiebrown01: After discussion it has been confirmed that we are not using `short_oucode` so the above is not a concern.

sophiebrown01 (approved):
Looks good, thanks for clarifying questions too! 🥇

PR #1525, opened 2026-05-06
ACD-1304: handle civil case proceedings with unknown application type codes

alexdesi:
## Summary

Civil case proceedings use application type codes **not currently** mapped in CDA's `supported_court_application_types.yaml`. 

Previously, unknown codes were silently misprocessed. The most dangerous case being incorrect representation orders submitted to HMCTS.

This PR makes CDA behave safely and visibly when encountering unknown codes:

- Removes the `BREACH_COURT_APPLICATIONS` feature flag. Breach applications are always enabled
- Removes `category_supported?` filtering so all court applications are returned regardless of category, including those with nil category (civil proceedings). This makes the application visible on VCD  search.
- Guards representation order creation with `supported_category?` — no Represenation Order is submitted to HMCTS for unknown categories, **preventing silent misrouting to the breach endpoint**
- Adds a `TODO` comment on judicial results for civil proceedings pending business confirmation
- Adds a `Rails.logger.warn` in `SupportedCourtApplicationTypes` when an unknown code is encountered, making these cases immediately visible in logs

### DIFF
--- a/app/models/hmcts_common_platform/court_application_summary.rb
+++ b/app/models/hmcts_common_platform/court_application_summary.rb
@@ -61,6 +61,8 @@ module HmctsCommonPlatform
       elsif application_category.in?(%w[breach poca])
         latest_hearing_judicial_results
       end
+      # TODO: Should judicial results be retrieved from hearings for civil case proceedings
+      # (nil category)? If so, the elsif above should include nil category or be removed entirely.
     end
 
     def subject_summary
@@ -75,10 +77,6 @@ module HmctsCommonPlatform
       # Warning: this `to_json` method doesn't return JSON, it returns a hash.
       # This is to be consistent with other models in this repo
       to_builder.attributes!
-    end
-
-    def category_supported?
-      true
     end
 
     def application_category
--- a/app/models/hmcts_common_platform/court_application_summary.rb
+++ b/app/models/hmcts_common_platform/court_application_summary.rb
@@ -78,11 +78,7 @@ module HmctsCommonPlatform
     end
 
     def category_supported?
-      if ENV["BREACH_COURT_APPLICATIONS"] == "true"
-        application_category.present?
-      else
-        application_category == "appeal"
-      end
+      application_category.present?
     end
 
     def application_category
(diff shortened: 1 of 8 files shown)


app/models/hmcts_common_platform/court_application_summary.rb
```diff
@@ -77,14 +79,6 @@ def to_json(*_args)
       to_builder.attributes!
     end
 
-    def category_supported?
-      if ENV["BREACH_COURT_APPLICATIONS"] == "true"
-        application_category.present?
-      else
-        application_category == "appeal"
-      end
-    end
```
alexdesi, line 86: We do not need this anymore, all the category are supported,
even those empty (the Civil cases too)

app/models/prosecution_case.rb
```diff
@@ -46,7 +46,6 @@ def load_hearing_results(defendant_id, load_events: true)
 
   def court_applications
     (body["applicationSummary"] || []).map { retrieve_full_court_application(it) }
-                                      .select(&:category_supported?)
```
alexdesi, line 49: this allow the endpoint to return the civil cases too.

spec/models/hmcts_common_platform/court_application_summary_spec.rb
```diff
@@ -50,28 +49,6 @@
     end
   end
 
-  describe "#category_supported?" do
-    subject(:supported) { court_application_summary.category_supported? }
-
-    context "when type is breach" do
-      let(:data) { { "applicationType" => "SE20521" } }
```
sophiebrown01, line 57: Just for my understanding, is this a random offence code or a criminal offence code? Making sense of the test

sophiebrown01: Actually don't worry, just realised you've removed this.

spec/models/court_application_spec.rb
```diff
@@ -14,4 +14,18 @@
 
     it { expect(court_application.appeal?).to be(false) }
   end
+
+  describe "#supported_category?" do
+    context "when the code maps to a known category" do
+      let(:application_type) { "MC80801" }
```
sophiebrown01, line 20: So this one is a known criminal offence code to verify?

PR #1526, opened 2026-05-08
Point UAT to HMCTS dev for civil proceedings testing

sophiebrown01:
This PR points CDA UAT to the HMCTS dev environment which they are using specifically for civil proceedings testing.


alexdesi (approved):
🚀 ✅

PR #1528, opened 2026-05-11
ACD-1351: Create errors CSV output for the import Xhibit

alexdesi:
Improves the XHIBIT cases import task to write failed rows to a CSV errors file.

- Adds CSV errors file output (`/tmp/import-errors.csv`) with all failed rows and their validation messages
- Makes `defendant_date_of_birth` optional with strict `yyyy-mm-dd` format validation when present
- Removes mandatory `defendant_arrest_summons_number` presence validation

### DIFF
--- a/app/services/import_xhibit_cases.rb
+++ b/app/services/import_xhibit_cases.rb
@@ -6,13 +6,13 @@ class ImportXhibitCases < ApplicationService
   end
 
   def call
-    results = { inserted: [], errors: [] }
-    CSV.foreach(file_path, headers: true).with_index(2) do |row, line|
+    results = { success_count: 0, errors: [] }
+    CSV.foreach(file_path, headers: true).with_index(2) do |row, line_number|
       xhibit_case = XhibitMigratedCase.create(row.to_h.transform_values(&:presence))
       if xhibit_case.persisted?
-        results[:inserted] << { line:, case_urn: xhibit_case.case_urn }
+        results[:success_count] += 1
       else
-        results[:errors] << { line:, case_urn: row["case_urn"], messages: xhibit_case.errors.full_messages }
+        results[:errors] << { line_number:, case_urn: row["case_urn"], row: row.to_h, messages: xhibit_case.errors.full_messages }
       end
     end
     results
--- a/lib/tasks/xhibit_cases.rake
+++ b/lib/tasks/xhibit_cases.rake
@@ -1,3 +1,5 @@
+require "csv"
+
 namespace :xhibit_cases do
   desc "Import XHIBIT cases from a CSV file. Example: rake xhibit_cases:import FILE_PATH=spec/fixtures/files/xhibit_cases_import.csv"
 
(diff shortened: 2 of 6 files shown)


lib/tasks/xhibit_cases.rake
```diff
@@ -12,11 +14,23 @@ namespace :xhibit_cases do
...
+
+    errors_file = "/tmp/import-errors.csv"
+    input_headers = CSV.open(file_path, headers: true) { |csv| csv.first&.headers } || []
+
+    CSV.open(errors_file, "w") do |csv|
+      csv << input_headers + %w[errors]
+
+      results[:errors].each do |error|
+        csv << error[:row].values_at(*input_headers) + [error[:messages].join(" | ")]
+      end
```
sophiebrown01: I was going to suggest that we perhaps make this conditional based on there being any errors, but then I am also happy for us to create an empty errors CSV per run if there turn out to be no errors.

app/services/import_xhibit_cases.rb
```diff
@@ -6,13 +6,13 @@ def initialize(file_path:)
...
-    CSV.foreach(file_path, headers: true).with_index(2) do |row, line|
+    results = { success_count: 0, errors: [] }
+    CSV.foreach(file_path, headers: true).with_index(2) do |row, line_number|
       xhibit_case = XhibitMigratedCase.create(row.to_h.transform_values(&:presence))
       if xhibit_case.persisted?
-        results[:inserted] << { line:, case_urn: xhibit_case.case_urn }
+        results[:success_count] += 1
       else
-        results[:errors] << { line:, case_urn: row["case_urn"], messages: xhibit_case.errors.full_messages }
+        results[:errors] << { line_number:, case_urn: row["case_urn"], row: row.to_h, messages: xhibit_case.errors.full_messages }
```
sophiebrown01, line 15: Just a note that we might not have a case urn (as it's not mandatory)

sophiebrown01:
I don't think I see it in this PR - we should also have a test to ensure that both the sent date and committal date are not empty. We also do not expect the sent date and committal date to _both_ be populated, this should be another error case. (so basically, the only valid condition is `sent_date XOR committal_date`

sophiebrown01 (approved):
Thanks for addressing those comments Alessandro, looks good! 🥇

lib/tasks/xhibit_cases.rake
alexdesi: this made me think we can timestamp the filename, so we know when it's been run, 
being empty is proof that no errors were present for that time.

`/tmp/import-errors-1780583810.csv` for example.

sophiebrown01:
I've done a check against the ACs for this task and I think there are a few missing tests, for:

- AC2: Duplicate URN
- AC9: Missing Case Type

I have referred back to the original DLRM requirements doc and it's worth noting that both case sub-type _and_ mode of trial are not mandatory. I can see in the original ACs that this distinction wasn't made, but if we receive a file without these fields then it is expected behaviour.

app/services/import_xhibit_cases.rb
alexdesi: Ok, I removed the `urn` mandatory, but we keep it the uniqueness (scoped within the defendant name and lastname)

alexdesi:
> I've done a check against the ACs for this task and I think there are a few missing tests, for:
> 
>     * AC2: Duplicate URN
Test was already present. 
It was added in the previous PR, so probably you missed it:
https://github.com/ministryofjustice/laa-court-data-adaptor/blob/main/spec/models/xhibit_migrated_case_spec.rb#L27

After you comment, I've improved the error message to be more explicit.
> 
>     * AC9: Missing Case Type
test already present: https://github.com/ministryofjustice/laa-court-data-adaptor/pull/1528/changes#diff-8855db9e6477e81c9c214840a6047b0873eac6ad875bb495095043e41c5c6f1bR29

> 
> 
> I have referred back to the original DLRM requirements doc and it's worth noting that both case sub-type _and_ mode of trial are not mandatory. I can see in the original ACs that this distinction wasn't made, but if we receive a file without these fields then it is expected behaviour.

Yes, as you said, case sub-type and mode of trial are not mandatory.
I've updated the code.

spec/services/import_xhibit_cases_spec.rb
```diff
@@ -17,8 +17,86 @@
...
+    end
+
+    it "reports a missing OU code as an error" do
+      result = import
+      error = result[:errors].find { |e| e[:case_urn] == "30GD0001003" }
+      expect(error).to include(line_number: 8, case_urn: "30GD0001003")
+      expect(error[:messages]).to include("Ou code can't be blank")
+    end
+
+    it "reports a missing case sub-type as an error" do
```
sophiebrown01: I think this should be case type as case sub-type is optional

spec/services/import_xhibit_cases_spec.rb
```diff
@@ -17,8 +17,86 @@
...
+      result = import
+      error = result[:errors].find { |e| e[:case_urn] == "30GD0001006" }
+      expect(error).to include(line_number: 11, case_urn: "30GD0001006")
+      expect(error[:messages]).to include("Case sub type can't be blank")
```
sophiebrown01: And this changed to be case type too

spec/services/import_xhibit_cases_spec.rb
alexdesi: yes, this is not needed anymore (and actually this was failing)

spec/services/import_xhibit_cases_spec.rb
alexdesi: not needed, removed

PR #1540, opened 2026-05-22
ACD-1419: replace synchronous rake task with async Sidekiq worker

alexdesi:
The original rake task was basically a script that ran everything in sequence, one MAAT ID at a time, sleeping 12 seconds between each one. For a big list this could
take hours — and if something crashed halfway through, you'd have no idea where it stopped or what had already been processed.

The new version is much simpler to run: you point it at a CSV file, it queues up a background job for each MAAT ID (staggered 12 seconds apart so we don't hammer the
HMCTS API), and exits immediately. Each job then runs independently via Sidekiq.

The other big improvement is that now every attempt — success or failure — gets saved to the database in maat_hearing_fetch_results. Before, results were just printed
to the terminal and lost. Now you can query them afterwards and see exactly what happened to each MAAT ID.

### DIFF
--- a/lib/tasks/hearings.rake
+++ b/lib/tasks/hearings.rake
@@ -1,48 +1,17 @@
 namespace :hearings do
-  desc "Fetch past hearings from HMCTS Common Platform for the given string of comma-separated MAAT IDs. \
-        Example: rake hearings:fetch '1234, 5678'"
-  task fetch: :environment do |_task, _args|
-    maat_ids = ARGV.last.split(",")
-    processed_maat_ids = []
-    errored_maat_ids = []
+  desc "Fetch past hearings from HMCTS Common Platform for MAAT IDs listed in a CSV file (one per row, no header). \
+        Example: rake 'hearings:fetch[path/to/maat_ids.csv]'"
+  task :fetch, [:csv_path] => :environment do |_task, args|
+    require "csv"
+    maat_ids = CSV.read(args[:csv_path]).flatten.map(&:strip)
 
-    puts "[INFO - #{Time.zone.now}] About to process #{maat_ids.count} MAAT IDs ..."
+    puts "[INFO - #{Time.zone.now}] Scheduling #{maat_ids.count} MAAT IDs ..."
 
-    maat_ids.count.times do
-      maat_id = maat_ids.shift
-
-      puts "\n\n"
-      puts "[INFO - #{Time.zone.now}] Processing #{maat_id} ..."
-
-      begin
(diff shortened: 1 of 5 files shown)


lib/tasks/hearings.rake
```diff
@@ -1,48 +1,17 @@
...
-        puts "[INFO - #{Time.zone.now}] Skipping MAAT ID #{maat_id}."
-        next
-      end
-
-      puts "[INFO - #{Time.zone.now}] Successfully processed MAAT ID #{maat_id}."
-      processed_maat_ids << maat_id
-
-      sleep 12
+    maat_ids.each_with_index do |maat_id, i|
+      MaatIdHearingFetcherWorker.perform_in(i * 12.seconds, maat_id)
```
alexdesi, line 11: this is the single most important line in this PR.

It does enqueue everything, but it does the call 12-seconds spaced.

sophiebrown01:
> The original rake task was basically a script that ran everything in sequence, one MAAT ID at a time, sleeping 12 seconds between each one. For a big list this could take hours — and if something crashed halfway through, you'd have no idea where it stopped or what had already been processed.
> 
> The new version is much simpler to run: you point it at a CSV file, it queues up a background job for each MAAT ID (staggered 12 seconds apart so we don't hammer the HMCTS API), and exits immediately. Each job then runs independently via Sidekiq.
> 
> The other big improvement is that now every attempt — success or failure — gets saved to the database in maat_hearing_fetch_results. Before, results were just printed to the terminal and lost. Now you can query them afterwards and see exactly what happened to each MAAT ID.

This looks great! I have a few thoughts:

- it may be required that we run this outside of core business hours, which could be added in as part of the enqueuing logic but would certainly complicate things. I think for the current re-pull exercise we only enqueue the MAAT ids that we know should be processed outside of business hours - if we have more then they can be processed the following day. Longer term though it is a consideration.
- You mentioned in the description that we're saving success or failure to a new DB table as part of this PR in `maat_hearing_fetch_results` but I don't see that in the changes - has that not been added yet? 🤔

alexdesi:
@sophiebrown01 this has been tested, 
even with big CSV file  4400+ entries.

Everything looks good 🟢 

I also added a log message here: https://github.com/ministryofjustice/laa-court-data-adaptor/pull/1540/changes#diff-61975489235f1101a88bd2541d3fce9a6d43506eb4cbe6e3be32420a3239233eR17

I am happy to merge it if you are.

sophiebrown01 (approved):
Great work Alessandro, looks good! 🚀

PR #1545, opened 2026-06-05
ACD-1437: report non-2xx CP responses to Sentry

alexdesi:
Adds Sentry observability for failed representation order calls to Common Platform without halting execution.

When CP returns a non-2xx response, `RecordProsecutionCaseRepresentationOrder` now captures the error via `Sentry.capture_exception` using a new `CommonPlatform::Api::Errors::UnsuccessfulHttpResponse` error class, then continues to update the local DB record as before.

### DIFF
--- a/app/services/common_platform/api/errors/unsuccessful_http_response.rb
+++ b/app/services/common_platform/api/errors/unsuccessful_http_response.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module CommonPlatform::Api::Errors
+  class UnsuccessfulHttpResponse < StandardError
+    # Common Platform was reached but returned a non-2xx HTTP status.
+    def initialize(service:, status:, body: nil)
+      super("#{service} failed with status #{status}: #{body}")
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/services/common_platform/api/errors/unsuccessful_http_response.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module CommonPlatform::Api::Errors
+  class UnsuccessfulHttpResponse < StandardError
+    # Common Platform was reached but returned a non-2xx HTTP status.
+    def initialize(service:, status:, body: nil)
+      super("#{service} failed with status #{status}: #{body}")
```
sophiebrown01, line 7: Just to double-check, the body will never contain any data from the request that failed?

sophiebrown01 (commented):
This looks good, with the caveat that this is only as a means of observability; the exceptions captured in Sentry will be non-actionable.

app/services/common_platform/api/errors/unsuccessful_http_response.rb
alexdesi: No, in case of error it returns something something like this:
```
{ "statusCode" => 401, "message" => "Access denied due to invalid subscription key. Make sure to provide a valid key for an active subscription." }
```

sophiebrown01: Thanks for confirming 👍

sophiebrown01 (approved):
Looks good 🥇

PR #1558, opened 2026-07-08
Add new Common Platform certificate to nginx config

sophiebrown01:
This PR adds the new Common Platform certificate (to replace the current certificate expiring this month) thumbprint to the nginx config as part of request authentication. Once connection is confirmed using the new certificate details, a separate PR will be raised to remove the old certificate thumbprint.

Interestingly in nginx config, you cannot have an `if` statement with multiple conditions; each has to be a separate `if` statement, with a common approach being to set a flag in these statements before a final `if` statement to test whether the flag has been set.

Reading:
- https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if
- https://medium.com/@prakharkhanna4/using-multiple-conditions-in-nginx-0f30a61eec21

[Link to story](https://dsdmoj.atlassian.net/browse/LAACONNECT-1473)


alexdesi (approved):
✅

PR #1563, opened 2026-07-10
Define pod resource requests and limits

sophiebrown01:
This PR defines the pod resource requests and limits for both CPU and memory. Previously these were not defined and would default to the following values:

CPU: request `75m`, limit `1`
Memory: request `750Mi`, limit `1000Mi`

CDA pods have recently been seeing CPU requests exceeding the request limit of `75m`, causing the pods to restart. This is likely due to overall load on CDA and needs further investigation, but in the meantime this commit ups the CPU request limit in an attempt to stabilise pods during increased workload.


alexdesi (approved):
it looks good to me! ✅

PR #1564, opened 2026-07-14
ACD-1472: Improve Common Platform logging

alexdesi:
When a call to HMCTS Common Platform failed, CDA raised generic errors ("Error posting LAA Reference to Common Platform") and logged only the HTTP status, discarding what Common Platform actually said. 

Every Common Platform call is now clearly identifiable in the logs with its duration, and failures carry the full detail of the response.

### DIFF
--- a/app/services/common_platform/connection.rb
+++ b/app/services/common_platform/connection.rb
@@ -14,12 +14,13 @@ module CommonPlatform
       @connection = Faraday.new HOST, options do |connection|
         connection.request :retry, retry_options
         connection.request :json
-        connection.response :logger, TaggedLogger, { headers: false } do |logger|
+        connection.response :logger, TaggedLogger, { headers: false, formatter: LogFormatter } do |logger|
           logger.filter(/(defendantName=)([^&]+)/, '\1[FILTERED]')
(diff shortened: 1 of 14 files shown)


sophiebrown01 (approved):
This is amazing! 🚀

app/services/common_platform/connection.rb
```diff
@@ -38,6 +38,31 @@ def call
...
 
+    # All Common Platform logging lives here.
+    # A request and a response line (with duration) for every call.
+    class LogFormatter < Faraday::Logging::Formatter
+      MAX_BODY_LENGTH = 500
+
+      def request(env)
+        env[:started_at] = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+        info { "Common Platform request: #{env.method.to_s.upcase} #{apply_filters(env.url.to_s)}" }
```
sophiebrown01, line 49: Just one question: when we have lots of requests in the logs, how will we know which info lines match for the request and the response times? If there's nothing to use in the request (like a correlation id or request id) then maybe we should create a random identifier to log in both the request and response output so we can visually match them?

alexdesi: Good question, yes, there is already a `Request ID`, provided by the class TaggedLogger.

PR #1567, opened 2026-07-17
ACD-1480: Fix undefined defendant_id

alexdesi:
When recording a representation order on Common Platform failed, reporting the error to Sentry crashed with an undefined `defendant_id`, masking the original failure.

The service now stores the defendant ID, so failures are reported to Sentry with the defendant and offence IDs.

### DIFF
--- a/app/services/common_platform/api/record_prosecution_case_representation_order.rb
+++ b/app/services/common_platform/api/record_prosecution_case_representation_order.rb
@@ -14,6 +14,7 @@ module CommonPlatform
                      effective_end_date: nil,
                      connection: CommonPlatform::Connection.instance.call)
         @case_defendant_offence = case_defendant_offence
+        @defendant_id = defendant_id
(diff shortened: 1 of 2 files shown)


app/services/common_platform/api/record_prosecution_case_representation_order.rb
```diff
@@ -14,6 +14,7 @@ def initialize(case_defendant_offence:,
                      effective_end_date: nil,
                      connection: CommonPlatform::Connection.instance.call)
         @case_defendant_offence = case_defendant_offence
+        @defendant_id = defendant_id
```
alexdesi, line 17: @sophiebrown01 this is the fix:
it stores `defendant_id` in an instance var, to be used later in the report, in case of CP failure.

sophiebrown01 (approved):
Thanks for fixing this Alessandro 🚀

PR #1568, opened 2026-07-17
ACD-1481: Fix NoMethodError for rep orders with no dated offences

alexdesi:
Sentry reported `NoMethodError: undefined method 'delete' for nil` in `RepresentationOrderCreator` when recording a representation order for a Breach/POCA court application.

With no granted representation order there is nothing to record on Common Platform, so the service now skips the call, matching the existing behaviour of the appeal and prosecution-case paths.

### DIFF
--- a/app/services/common_platform/api/representation_order_creator.rb
+++ b/app/services/common_platform/api/representation_order_creator.rb
@@ -27,6 +27,8 @@ module CommonPlatform
       end
 
       def call
+        return if offences.empty?
+
         call_common_platform_endpoint
       end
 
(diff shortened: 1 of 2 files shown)


app/services/common_platform/api/representation_order_creator.rb
```diff
@@ -27,6 +27,8 @@ def initialize(defendant_id:, offences:, maat_reference:, defence_organisation:)
       end
 
       def call
+        return if offences.empty?
```
alexdesi, line 30: this is so a payload whose offences all lack a status_date is skipped instead of reaching the Breach/POCA branch and crashing on offences.first.

sophiebrown01 (approved):
Looks good 🥇

PR #1569, opened 2026-07-20
ACD-1474: Update the bundle and fix warnings

alexdesi:
- Updates Rails from 8.0.3 to 8.1.3, along with other dependencies, keeping the application on a supported, patched version of the framework.
- Fixes the warnings surfaced by the Rails 8.1 upgrade and by Brakeman, with no behavioural change for callers of the affected endpoint.


sophiebrown01 (approved):
Looks good! I think we should definitely test this with the end to end tests by pointing a local VCD to one of the environments once these changes are deployed there but all makes sense 🚀

PR #1577, opened 2026-07-24
ACD-1484: Handle slow Common Platform responses

alexdesi:
When Common Platform responses stall, API pods run out of NGINX threads. The `/status` health check shares that same thread pool, so it is starved of a thread and Kubernetes restarts the pod, removing capacity at exactly the moment Common Platform is already degraded, and shifting load onto the remaining pods.

For caseworkers this appears as hanging or failed court data requests during Common Platform slowdowns, made worse by pods cycling out of service.

This bounds how long a single Common Platform call can hold a thread, and doubles the
threads available per pod:

- `open_timeout: 3`: connection setup was previously unbounded and fell back to
  Net::HTTP's 60s default, so the existing 30s read timeout never applied when Common
  Platform stopped completing handshakes
- `read_timeout: 10` (was 30): threads are released three times faster when a call stalls
- `RAILS_MAX_THREADS: 10` (was 5):  doubles concurrent request capacity per pod


helm_deploy/laa-court-data-adaptor/templates/_environment.tpl
```diff
@@ -16,6 +16,8 @@ env:
         key: url
   - name: RAILS_ENV
     value: 'production'
+  - name: RAILS_MAX_THREADS
```
alexdesi, line 19: These were not specified, so it defaulted to only 5 connections

app/services/common_platform/connection.rb
```diff
@@ -27,7 +27,8 @@ def initialize
           keep_alive: 60,
           pool_size: 10,    # to safetly handle For 3-5 req/sec
           idle_timeout: 120,
-          read_timeout: 30, # 30 seconds: to have safe buffer for slow responses
+          open_timeout: 3,  # connect + TLS only; without this Net::HTTP defaults to 60 seconds
+          read_timeout: 10,
```
alexdesi, line 31: These are to release the connection more quickly when it does not respond.

sophiebrown01 (approved):
Looks good 👍

PR #1578, opened 2026-07-27
Add the CP url to spot the URL involved in the slow responses.

alexdesi:
## What
Add the CP url to spot the URL involved in the slow responses.


sophiebrown01 (approved):
Fantastic, thanks for this! Will give us better visibility 🚀

PR #1580, opened 2026-07-29
ACD-1483: Report to Sentry when a defendant cannot be linked

alexdesi:
When a caseworker tries to link a defendant to a MAAT ID, the request is rejected if
Common Platform has not provided any hearings for the case. 

These failures are now reported to Sentry and to the application log, making it possible to spot how often this happens and which cases are affected, and to raise the missing data with HMCTS.

No change to what is returned to the client: linking behaves exactly as before.


sophiebrown01 (approved):
Thanks for this! Good to know about these edge cases if they happen 💯

PR #1582, opened 2026-07-31
ACT-1355: Search MAAT applications in MAAT

alexdesi:
Adds a service that asks MAAT whether a representation order already exists for a defendant, and whether it is already linked to a case.

This is the first step towards automatically linking defendants to their MAAT application.


leandroalemao (approved):
👍 simple, clean and easy to follow ✅ tks @alexdesi

PR #1585, opened 2026-08-04
ACD-1356: Process pending XHIBIT migrated cases

alexdesi:
XHIBIT cases imported into the adaptor sit in a **pending state** , CDA now asks MAAT whether an application exists for the defendant's name.

When MAAT has no match, the case is flagged as `action_required` so a caseworker can pick it up.


app/services/process_xhibit_cases.rb
```diff
@@ -0,0 +1,34 @@
...
+      last_name: xhibit_case.defendant_last_name,
+    )
+
+    case response&.status
+    when SUCCESS
+      # TODO: process the maat applications matching the case.
+    when NOT_FOUND
+      xhibit_case.action_required!
+    else
+      record_error(xhibit_case, status: response&.status, body: response&.body)
```
leandroalemao: Hi @alexdesi How about recording an error using the same key names to keep it consistent with when a `Faraday::Error` happens?

`record_error(xhibit_case, error: response&.status, message: response&.body)`

Is it possible to add a spec for this scenario as well please?

alexdesi: Yes, that's a good idea.

app/services/process_xhibit_cases.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+  end
+
+private
+
+  def process(xhibit_case)
+    response = MaatApi::MaatApplicationSearcher.call(
+      first_name: xhibit_case.defendant_first_name,
+      last_name: xhibit_case.defendant_last_name,
+    )
```
sophiebrown01: To check, is this an initial pass as calling MAAT API to find a MAAT application for a given migrated Xhibit case? I ask this as we want to search on as many of the fields as we have been supplied by HMCTS in the csv file - if we're only searching by first and last name then we may get many MAAT applications back which would need manual intervention when we could possibly narrow down to a single application if we provided all of the values that we can search on at this MAAT API endpoint.

spec/services/process_xhibit_cases_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+                     match_requests_on: %i[method uri body]) do
+      example.run
+    end
+  end
+
+  it "set status to (manual) action_required when there is no matching maat application" do
+    process_cases
+
+    expect(xhibit_case.reload).to be_action_required
+  end
```
pezholio: I'd probably expect this to be wrapped in a context too, just to match the scenarios when there's a matching and failed application. Would just make it a bit more readable.

pezholio: Ah, ignore me, this seems to be addressed in the next commit

pezholio (approved):
LGTM 👍

leandroalemao (approved):

PR #1588, opened 2026-08-07
Make setup easier

pezholio:
This helps address some of the issues I had getting set up, and allow easy setup talking to the hosted version of `hmcts-common-platform-mock-api` with minimal fuss. I may have made some assumptions about how things work, so any feedback, good or bad is appreciated!


leandroalemao (approved):
tested and looking good.. tkss 👍 😄

PR #1591, opened 2026-08-13
(LAACONNNECT-1357) Implement auto linking scenario for unlinked maat application

pezholio:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/LAACONNECT-1357)

## Checklist

This handles the "happy path" for auto-linking when the MAAT API returns one result for an imported case. If this is the case we:

- Call `CourtApplicationMaatLinkCreator` (which updates the MAAT ID in Common Platform and posts a link message to the SQS MAAT queue)
- Updates the xhbit case record, setting the `status` to `AUTO_LINKED`, `linked_date` to the current timestamp, and `linked_by `to `SYSTEM`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,3 +30,6 @@ RSpec/VariableName:
 RSpec/EmptyExampleGroup:
   Exclude:
     - 'spec/requests/**/*.rb'
+
+Rails/Delegate:
+  Enabled: false
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -16,11 +16,31 @@ module MaatApi
       response["maatId"]
     end
 
+    def no_existing_link?
+      !is_linked? && !libra_id && !case_urn
+    end
+
   private
 
     attr_reader :faraday_response
 
     def_delegators :faraday_response, :status, :success?
+
+    def libra_id
+      linking_detail["libraId"]
+    end
+
+    def case_urn
+      linking_detail["caseUrn"]
+    end
+
+    def is_linked?
+      response["isLinked"] == true
+    end
+
+    def linking_detail
+      response["linkingDetail"] || {}
+    end
 
     # Return the first item in the array by default, as this is the happy path
     # going forward, we'll handle the case of multiple results
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -1,11 +1,19 @@
 module MaatApi
   class SearchResponse
-    extend Forwardable
-
-    def_delegator :faraday_response, :body
-
     def initialize(faraday_response)
       @faraday_response = faraday_response
+    end
+
+    def body
+      faraday_response.body
+    end
+
+    def status
+      faraday_response.status
+    end
+
+    def success?
+      faraday_response.success?
     end
 
     def not_found?
@@ -23,8 +31,6 @@ module MaatApi
   private
 
     attr_reader :faraday_response
-
-    def_delegators :faraday_response, :status, :success?
 
     def libra_id
       linking_detail["libraId"]
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -1,19 +1,19 @@
 module MaatApi
   class SearchResponse
-    def initialize(faraday_response)
-      @faraday_response = faraday_response
+    def initialize(http_response)
+      @http_response = http_response
     end
 
     def body
-      faraday_response.body
+      http_response.body
     end
 
     def status
-      faraday_response.status
+      http_response.status
     end
 
     def success?
-      faraday_response.success?
+      http_response.success?
     end
 
     def not_found?
@@ -30,7 +30,7 @@ module MaatApi
 
   private
 
-    attr_reader :faraday_response
+    attr_reader :http_response
 
     def libra_id
       linking_detail["libraId"]
@@ -51,7 +51,7 @@ module MaatApi
     # Return the first item in the array by default, as this is the happy path
     # going forward, we'll handle the case of multiple results
     def response
-      faraday_response&.body&.first || {}
+      http_response&.body&.first || {}
     end
   end
 end
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -12,10 +12,20 @@ module MaatApi
       status == 404
     end
 
+    def maat_id
+      response["maatId"]
+    end
+
   private
 
     attr_reader :faraday_response
 
     def_delegators :faraday_response, :status, :success?
+
+    # Return the first item in the array by default, as this is the happy path
+    # going forward, we'll handle the case of multiple results
+    def response
+      faraday_response&.body&.first || {}
+    end
   end
 end
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -1,0 +1,21 @@
+module MaatApi
+  class SearchResponse
+    extend Forwardable
+
+    def_delegator :faraday_response, :body
+
+    def initialize(faraday_response)
+      @faraday_response = faraday_response
+    end
+
+    def not_found?
+      status == 404
+    end
+
+  private
+
+    attr_reader :faraday_response
+
+    def_delegators :faraday_response, :status, :success?
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  SYSTEM_USERNAME = "SYSTEM"
+
   has_secure_password :auth_token
 
   validates :name, presence: true
--- a/app/services/link_xhibit_case.rb
+++ b/app/services/link_xhibit_case.rb
@@ -1,0 +1,27 @@
+class LinkXhibitCase < ApplicationService
+  class LinkFailure < StandardError; end
+
+  def call(maat_search_response, xhibit_case)
+    create_maat_link!(xhibit_case.defendant_id, maat_search_response.maat_id)
+    update_xhibit_case!(xhibit_case, maat_search_response.maat_id)
+  end
+
+private
+
+  def create_maat_link!(defendant_id, maat_id)
+    CourtApplicationMaatLinkCreator.call(
+      defendant_id,
+      "SYSTEM",
+      maat_id,
+    )
+  end
+
+  def update_xhibit_case!(xhibit_case, maat_id)
+    xhibit_case.update!(
+      maat_id: maat_id,
+      status: :auto_linked,
+      linked_at: Time.zone.now,
+      linked_by: "SYSTEM",
+    )
+  end
+end
--- a/app/services/link_xhibit_case.rb
+++ b/app/services/link_xhibit_case.rb
@@ -1,27 +1,43 @@
 class LinkXhibitCase < ApplicationService
-  class LinkFailure < StandardError; end
+  TRIAL = "T".freeze
+  SENTENCE = "S".freeze
+  APPEAL = "A".freeze
+  USER_NAME = "SYSTEM".freeze
 
-  def call(maat_search_response, xhibit_case)
-    create_maat_link!(xhibit_case.defendant_id, maat_search_response.maat_id)
-    update_xhibit_case!(xhibit_case, maat_search_response.maat_id)
+  def initialize(maat_search_response, xhibit_case, court_data)
+    @maat_search_response = maat_search_response
+    @xhibit_case = xhibit_case
+    @court_data = court_data
+  end
+
+  def call
+    case xhibit_case.case_type
+    when TRIAL
+      ProsecutionCaseMaatLinkCreator.call(court_data.defendant_id, USER_NAME, maat_id)
+    when SENTENCE, APPEAL
+      CourtApplicationMaatLinkCreator.call(subject_id, USER_NAME, maat_id)
+    else
+      raise ArgumentError, "Unsupported case type: #{xhibit_case.case_type.inspect}"
+    end
+
+    xhibit_case.update!(
+      maat_id:,
+      status: :auto_linked,
+      linked_at: Time.zone.now,
+      linked_by: USER_NAME,
+    )
   end
 
 private
 
-  def create_maat_link!(defendant_id, maat_id)
-    CourtApplicationMaatLinkCreator.call(
-      defendant_id,
-      "SYSTEM",
-      maat_id,
-    )
+  attr_reader :maat_search_response, :xhibit_case, :court_data
+
+  def subject_id
+    court_data.application_summaries.first&.subject_summary&.subject_id ||
+      raise(ArgumentError, "No court application found for defendant #{court_data.defendant_id}")
   end
 
-  def update_xhibit_case!(xhibit_case, maat_id)
-    xhibit_case.update!(
-      maat_id: maat_id,
-      status: :auto_linked,
-      linked_at: Time.zone.now,
-      linked_by: "SYSTEM",
-    )
+  def maat_id
+    maat_search_response.maat_id
   end
 end
--- a/app/services/link_xhibit_case.rb
+++ b/app/services/link_xhibit_case.rb
@@ -2,7 +2,6 @@ class LinkXhibitCase < ApplicationService
   TRIAL = "T".freeze
   SENTENCE = "S".freeze
   APPEAL = "A".freeze
-  USER_NAME = "SYSTEM".freeze
 
   def initialize(maat_search_response, xhibit_case, court_data)
     @maat_search_response = maat_search_response
@@ -13,9 +12,9 @@ class LinkXhibitCase < ApplicationService
   def call
     case xhibit_case.case_type
     when TRIAL
-      ProsecutionCaseMaatLinkCreator.call(court_data.defendant_id, USER_NAME, maat_id)
+      ProsecutionCaseMaatLinkCreator.call(court_data.defendant_id, User::SYSTEM_USERNAME, maat_id, can_update_laa_status: true)
     when SENTENCE, APPEAL
-      CourtApplicationMaatLinkCreator.call(subject_id, USER_NAME, maat_id)
+      CourtApplicationMaatLinkCreator.call(subject_id, User::SYSTEM_USERNAME, maat_id, can_update_laa_status: true)
     else
       raise ArgumentError, "Unsupported case type: #{xhibit_case.case_type.inspect}"
     end
@@ -24,7 +23,7 @@ class LinkXhibitCase < ApplicationService
       maat_id:,
       status: :auto_linked,
       linked_at: Time.zone.now,
-      linked_by: USER_NAME,
+      linked_by: User::SYSTEM_USERNAME,
     )
   end
 
--- a/app/services/maat_api/maat_application_searcher.rb
+++ b/app/services/maat_api/maat_application_searcher.rb
@@ -23,7 +23,9 @@ module MaatApi
     end
 
     def call
-      @connection.presence&.post(URL, @search_request)
+      MaatApi::SearchResponse.new(
+        @connection.presence&.post(URL, @search_request),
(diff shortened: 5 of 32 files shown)


app/services/maat_api/maat_application_searcher.rb
```diff
@@ -23,7 +23,9 @@ def initialize(first_name: nil,
     end
 
     def call
-      @connection.presence&.post(URL, @search_request)
+      MaatApi::SearchResponse.new(
```
leandroalemao: Hi @pezholio do you think we need to rescue for some exceptions here in case the post request does not work? or Faraday is configured to not raise exceptions anyway so we don't need to worry about catching exceptions here? https://lostisland.github.io/faraday/#/middleware/included/raising-errors

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
...
+  class LinkFailure < StandardError; end
+
+  def call(maat_search_response, xhibit_case)
+    create_maat_link(xhibit_case.case_urn, maat_search_response.maat_id)
+    update_exhibit_case(xhibit_case, maat_search_response.maat_id)
+  end
+
+private
+
+  def create_maat_link(case_urn, maat_id)
```
leandroalemao: Hi @pezholio what do you think about adding `!` to these methods as they can raise exceptions? just for consistency? tks

leandroalemao (commented):
Hi @pezholio LGTM .. just left a small comment really.. tksss 👏 👍

app/services/maat_api/maat_application_searcher.rb
pezholio: I think this is already handled in `ProcessXhibitCases` here https://github.com/ministryofjustice/laa-court-data-adaptor/blob/7c30ad2df74ecf9daa145c973473140f1532eb5e/app/services/process_xhibit_cases.rb#L17

app/services/link_xhibit_case.rb
pezholio: Good plan 👍

spec/services/link_xhibit_case_spec.rb
```diff
@@ -0,0 +1,41 @@
+RSpec.describe LinkXhibitCase, type: :service do
+  describe "#call" do
+    let(:maat_response) { instance_double(MaatApi::SearchResponse, maat_id: "6559879") }
+    let(:xhibit_case) do
+      XhibitMigratedCase.create!(
```
leandroalemao: Hi @pezholio one for another ticket I guess.. should we add factory bot to the project? thanks

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
...
+  def call(maat_search_response, xhibit_case)
+    create_maat_link!(xhibit_case.case_urn, maat_search_response.maat_id)
+    update_exhibit_case!(xhibit_case, maat_search_response.maat_id)
+  end
+
+private
+
+  def create_maat_link!(case_urn, maat_id)
+    CourtApplicationMaatLinkCreator.call(
+      case_urn,
```
leandroalemao: Hi @pezholio think `CourtApplicationMaatLinkCreator` expects the `subject_id` (`defendant_id`) in the params instead?

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
+class LinkXhibitCase < ApplicationService
+  class LinkFailure < StandardError; end
+
+  def call(maat_search_response, xhibit_case)
+    create_maat_link!(xhibit_case.case_urn, maat_search_response.maat_id)
```
leandroalemao: Hi @pezholio should we wrap these in a `ActiveRecord::Base.transaction` block? in case any of the actions blows up, it then rolls back.

spec/services/link_xhibit_case_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        sent_date: Date.new(2019, 10, 25),
+      )
+    end
+
+    let(:link_case) { described_class.new.call(maat_response, xhibit_case) }
+
+    before do
+      allow(CourtApplicationMaatLinkCreator).to receive(:call)
+    end
+
```
leandroalemao, line 31: Hi @pezholio should we add a `failure path` to the specs? A test with a nil/blank `maat_id` maybe? If we add validations to `XhibitMigratedCase` model to make sure we have `maat_id`, `linked_by` and `linked_at`, when the status is in one the linked states maybe?

leandroalemao (commented):
Hi @pezholio I've just added a few more comments... tks 👍

spec/services/link_xhibit_case_spec.rb
pezholio: Oh I agree - was surprised to see it not here to be honest. Happy to add it here actually.

pezholio: Actually, will make it another PR 👍

app/services/link_xhibit_case.rb
pezholio: Ah, nice catch!

spec/services/link_xhibit_case_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        defendant_first_name: "Alice",
+        defendant_last_name: "Smith",
+        sent_date: Date.new(2019, 10, 25),
+      )
+    end
+
+    let(:link_case) { described_class.new.call(maat_response, xhibit_case) }
+
+    before do
+      allow(CourtApplicationMaatLinkCreator).to receive(:call)
```
leandroalemao, line 29: Hi @pezholio maybe if you do `allow(CourtApplicationMaatLinkCreator).to receive(:call).and_call_original`, the specs would have failed due the wrong param being sent to the `CourtApplicationMaatLinkCreator`

spec/services/link_xhibit_case_spec.rb
leandroalemao: Hi @pezholio yeah another PR for adding factory bot is a good plan. tks

app/services/link_xhibit_case.rb
pezholio: Yeah, I did wonder about that, but I was focussing on the happy path for this PR. That said, I think it's worth adding 👍

spec/services/link_xhibit_case_spec.rb
pezholio: Yeah, I think I'd end up having to do a bunch of mocking and setup further down the stack to get the tests working that way. I'd rather keep the unit tests nice and focussed here. It would be a good idea to have to more detailed functional tests though. I'll see where those might belong as part of this.

app/services/link_xhibit_case.rb
pezholio: Thinking about it, I don't think there's any value here, as there's only one database call here. If the call to `CourtApplicationMaatLinkCreator` blows up, the call to update the xhibit case won't call, so we're safe there. The only minor risk is if the call to update the xhibit case doesn't succeed, but I think if that happens, we'll have bigger problems.

spec/services/link_xhibit_case_spec.rb
pezholio: PR is here https://github.com/ministryofjustice/laa-court-data-adaptor/pull/1593

spec/cassettes/maat_api/search_maat_application_multiple_results.yml
```diff
@@ -0,0 +1,110 @@
...
+        Date:
+          - Fri, 31 Jul 2026 15:22:23 GMT
+        Content-Type:
+          - application/json
+        Transfer-Encoding:
+          - chunked
+        Connection:
+          - keep-alive
+        Set-Cookie:
+          - INGRESSCOOKIE=1785511343.432.31305.664596|55b9a069646b5e4eb9ec5bafd08dbaec;
```
sophiebrown01, line 90: This is probably fine but just checking...

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def call(maat_search_response, xhibit_case)
+    create_maat_link!(xhibit_case.defendant_id, maat_search_response.maat_id)
+    update_exhibit_case!(xhibit_case, maat_search_response.maat_id)
+  end
+
+private
+
+  def create_maat_link!(defendant_id, maat_id)
+    CourtApplicationMaatLinkCreator.call(
```
sophiebrown01: We will also need to update this call downstream (possibly passing the arg in here) to specify that the legal aid status is resubmitted through to CDA and updated back in Common Platform. I think this is missing from the task description referring back to it - let me try and find what the expected param is and get back to you on this.

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
...
+      defendant_id,
+      "SYSTEM",
+      maat_id,
+    )
+  end
+
+  def update_exhibit_case!(xhibit_case, maat_id)
```
sophiebrown01: Small typo, should be `xhibit` not `exhibit` 🙂

sophiebrown01 (commented):
Looks good Stuart! A few small points and then the wider discussion in our team channel, but that isn't as a result of this PR 🙂

spec/cassettes/maat_api/search_maat_application_multiple_results.yml
pezholio: To be honest, I just copied the cassette and changed the data. It looks like the other cassettes have the same headers in place. We could probably filter that info out when we record the cassettes though, as they're not relevant to the response.

app/models/maat_api/search_response.rb
```diff
@@ -0,0 +1,51 @@
...
+    def not_found?
+      status == 404
+    end
+
+    def maat_id
+      response["maatId"]
+    end
+
+    def no_existing_link?
+      !is_linked? && !libra_id && !case_urn
```
alexdesi, line 28: I am looking for clarification, and need to catch up with the logic:
is not enough `!is_linked?`

app/models/maat_api/search_response.rb
```diff
@@ -0,0 +1,51 @@
...
+    def libra_id
+      linking_detail["libraId"]
+    end
+
+    def case_urn
+      linking_detail["caseUrn"]
+    end
+
+    def is_linked?
+      response["isLinked"] == true
```
alexdesi, line 44: just checking, 
Is response["isLinked"] a boolean (or string)?

app/models/maat_api/search_response.rb
```diff
@@ -0,0 +1,51 @@
+module MaatApi
+  class SearchResponse
+    extend Forwardable
+
+    def_delegator :faraday_response, :body
```
alexdesi: @pezholio we usually tend to avoid `def_delegator` and where possible, be more explicit.
Even if it means being more wordy sometimes.

in this case, this line translates to:
```
def body
  faraday_response.body
end
```

which is Greppable, easier to read, no dependencies (Forwardable)

app/models/maat_api/search_response.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def no_existing_link?
+      !is_linked? && !libra_id && !case_urn
+    end
+
+  private
+
+    attr_reader :faraday_response
+
+    def_delegators :faraday_response, :status, :success?
```
alexdesi: and here:
```
private

def status
  faraday_response.status
end

def success?
  faraday_response.success?
end
```

(Double-check if we need them)

app/models/maat_api/search_response.rb
pezholio: The scenario mentions all three - https://dsdmoj.atlassian.net/wiki/spaces/ACD2/pages/6092227492/DLRM+Implementation+Analysis#Scenario-3%3A-MAAT-record-found%2C-no-existing-link

app/models/maat_api/search_response.rb
pezholio: The cassettes seem to suggest it comes out as a boolean

app/models/maat_api/search_response.rb
pezholio: That sounds sensible 👍

app/services/process_xhibit_cases.rb
```diff
@@ -1,7 +1,4 @@
 class ProcessXhibitCases < ApplicationService
-  NOT_FOUND = 404
-  SUCCESS = 200..299
```
alexdesi, line 3: your refactor looks okay, but double check it.
if I am not wrong, when I tested the API,
there were some scenarios that needed this trick.

app/services/process_xhibit_cases.rb
```diff
@@ -30,18 +27,19 @@ def maat_search(xhibit_case)
...
       handle_not_found(xhibit_case)
     else
       record_error(xhibit_case, error: response&.status, message: response&.body)
     end
   end
 
-  def handle_success(_response, _xhibit_case)
-    # TODO: process the maat applications matching the case.
+  def handle_success(response, xhibit_case)
+    LinkXhibitCase.call(response, xhibit_case) if response.no_existing_link?
```
alexdesi: if it has already been linked is an important to log it,
a silent failure is difficult to debug

pezholio: We've got some more tickets to handle the unhappy paths, but I'll add some logging for now 👍

spec/models/maat_api/search_response_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+    it "returns true when is_linked is present but not strictly true" do
+      faraday_response = instance_double(Faraday::Response, status: 200, body: [
+        { "isLinked" => "yes", "linkingDetail" => {} },
+      ])
+      response = described_class.new(faraday_response)
+
+      expect(response.no_existing_link?).to be true
+    end
+  end
+end
```
alexdesi, line 193: @pezholio these tests are really thorough, and a very fine-grained level, 
which is not actually bad, 
but to make it more readable we should focus on business logic on a real (fixture json) instead of using tons of instance_double.

What do you think?

alexdesi: some of these tests look like we are testing faraday (but the class is a model `SearchResponse` , we should not care too much on the http client)

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
+class LinkXhibitCase < ApplicationService
+  class LinkFailure < StandardError; end
```
alexdesi: is this LinkFailure error raised somewhere? 🤔

app/services/link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
...
+    create_maat_link!(xhibit_case.defendant_id, maat_search_response.maat_id)
+    update_xhibit_case!(xhibit_case, maat_search_response.maat_id)
+  end
+
+private
+
+  def create_maat_link!(defendant_id, maat_id)
+    CourtApplicationMaatLinkCreator.call(
+      defendant_id,
+      "SYSTEM",
```
alexdesi: is this string deliberate? 🤔 
it sounds we should use a constant for this magic string, 
maybe something like 
const SYSTEM_USER = "SYSTEM"
?

app/services/link_xhibit_case.rb
pezholio: Ha, whoops! That was left over from a previous refactor

app/services/link_xhibit_case.rb
pezholio: Good plan 👍

spec/models/maat_api/search_response_spec.rb
pezholio: I think it's probably OK to use a double for this class, and lean on the fixtures for tests that are more high level (such as the process_xhibit_cases_spec). That said, these tests could be a bit more readable. I'll have a crack :+1:

app/services/link_xhibit_case.rb
pezholio: Have addressed this now - see the last three commits 👍

app/services/process_xhibit_cases.rb
```diff
@@ -1,20 +1,22 @@
 class ProcessXhibitCases < ApplicationService
   def call
     XhibitMigratedCase.pending.find_each do |xhibit_case|
-      process_case(xhibit_case)
```
pezholio, line 7: This class is getting a bit complex now, with a lot of passing around of `response` and `xhibit_case` variables. I wonder it it'd be easier to have a `ProcessXhibitCase` class that accepts a single `xhibit_case` argument, which might be easier to work with?

app/services/link_xhibit_case.rb
```diff
@@ -1,27 +1,43 @@
...
-  def create_maat_link!(defendant_id, maat_id)
-    CourtApplicationMaatLinkCreator.call(
-      defendant_id,
-      "SYSTEM",
-      maat_id,
-    )
-  end
+  def call
+    case xhibit_case.case_type
+    when TRIAL
```
pezholio, line 14: I like this!

app/services/link_xhibit_case.rb
```diff
@@ -1,27 +1,43 @@
 class LinkXhibitCase < ApplicationService
-  class LinkFailure < StandardError; end
+  TRIAL = "T".freeze
+  SENTENCE = "S".freeze
+  APPEAL = "A".freeze
+  USER_NAME = "SYSTEM".freeze
```
pezholio: I wonder whether `USER_NAME` should be within the user class, ie: `User::SYSTEM_USERNAME`

.rubocop.yml
```diff
@@ -30,3 +30,6 @@ RSpec/VariableName:
 RSpec/EmptyExampleGroup:
   Exclude:
     - 'spec/requests/**/*.rb'
+
+Rails/Delegate:
+  Enabled: false
```
pezholio, line 35: 👍

app/models/maat_api/search_response.rb
```diff
@@ -1,19 +1,19 @@
 module MaatApi
   class SearchResponse
-    def initialize(faraday_response)
-      @faraday_response = faraday_response
+    def initialize(http_response)
```
pezholio, line 3: This is much better naming!

pezholio (commented):
Looking good - just realised I can't approve, but have made a couple of comments regardless 👍

app/services/process_xhibit_cases.rb
alexdesi: I see what you mean, 
I think it's still managable, 
let's see if became more complex building other tickets, in case we'll refactor.
Also I like I can read and reason on a single place instead of splitting it into many mini 3 lines methods.

app/services/link_xhibit_case.rb
alexdesi: yes, I agree

app/models/user.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
+  SYSTEM_USERNAME = "SYSTEM"
```
pezholio, line 4: 👍

spec/services/xhibit_auto_linking_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+      expect(xhibit_case.reload).to be_auto_linked
+      expect(xhibit_case).to have_attributes(
+        maat_id: maat_id.to_s,
+        linked_by: User::SYSTEM_USERNAME,
+        linked_at: within(1.minute).of(Time.zone.now),
+      )
+    end
+
+    it "publishes one MAAT link message with LAA status reinstatement enabled" do
+      expect(Sqs::MessagePublisher).to receive(:call).once.with(
```
pezholio: I tend to prefer using `have_received` after the code under test has been called, particularly as `Sqs::MessagePublisher#call` has already been stubbed. It just reads better to my eyes

alexdesi: yes, it makes sense!

leandroalemao (approved):

PR #1593, opened 2026-08-14
Use factory bot in tests

pezholio:
As suggested in https://github.com/ministryofjustice/laa-court-data-adaptor/pull/1591#discussion_r3782495486, this adds FactoryBot as a dependency to reduce the friction involved in creating objects in tests.


leandroalemao (approved):
👍 👏 tkssss

PR #1594, opened 2026-08-17
(LAACONNECT-1498) Route LAA Status Update API Calls Based on With/Without Offences

pezholio:
[Link to story](https://dsdmoj.atlassian.net/browse/LAACONNECT-1498)

Under the original agreement with HMCTS:

- Appeal applications were expected to always contain offences.
- Breach and POCA applications were expected to never contain offences.

Based on these business rules, we used the application type to determine LAA Status Update API should be called.

However, during end-to-end testing of DLRM-migrated cases, we identified a Breach application that contained offence data. HMCTS subsequently confirmed that exceptions to the original business rules were required to support the migration of cases from XHIBIT to Common Platform.

This PR therefore changes the logic to inspect the body of an application - if it contains offences, we loop through the offences we have been given by MLRA and make a call to Common Platform for each offence. If it does not contain offences, we know that the offence we have been given is a dummy offence, and we use this to build the payload and call the API endpoint without offences.

### DIFF
--- a/app/models/court_application.rb
+++ b/app/models/court_application.rb
@@ -9,6 +9,10 @@ class CourtApplication < ApplicationRecord
     body["hearingSummary"]
   end
 
+  def has_offences?
+    body.dig("subjectSummary", "offenceSummary")&.present?
+  end
+
   def appeal?
     application_type = body["applicationType"]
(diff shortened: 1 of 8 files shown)


leandroalemao (approved):
lgtm 👍 😄 ✅  tksss

PR #1597, opened 2026-08-21
ACD-1508: alert when a court centre code cannot be resolved

alexdesi:
This adds visibility, **not a fix**: 

- a Sentry warning when Common Platform omits the code
- and a Sentry error when it cannot be resolved from our reference data either.


leandroalemao (commented):
Hi @alexdesi is there any specs you'd add please? tksss

leandroalemao (approved):
Hi @alexdesi tksss 👍 😄

PR #1600, opened 2026-09-07
1358: implement auto-linking sceneario for MAAT application linked to a common platform case

alexdesi:
The specific implementation details for this scenario can be found in the documented technical implementation ([DLRM Implementation Analysis | Scenario 2: MAAT record found, already linked to a Common Platform case](https://dsdmoj.atlassian.net/wiki/spaces/ACD2/pages/6092227492/DLRM+Implementation+Analysis#Scenario-2%3A-MAAT-record-found%2C-already-linked-to-a-Common-Platform-case) ).

To summarise what this says:

    update the corresponding row in the xhibit_migrated_cases table with the following information:

        status: MANUAL_ACTION_REQUIRED

        error_message: MAAT ID already linked with other CP case ({caseUrn})

A MAAT application can be determined to be linked to an existing Common Platform case so long as the libraId field contains the CP prefix and the caseUrn field is not null or empty.

### DIFF
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -1,5 +1,7 @@
 module MaatApi
   class SearchResponse
+    COMMON_PLATFORM_PREFIX = "CP".freeze
+
     def initialize(http_response)
       @http_response = http_response
     end
@@ -24,8 +26,16 @@ module MaatApi
       response["maatId"]
     end
 
-    def no_existing_link?
-      !is_linked? && !libra_id && !case_urn
+    def existing_link?
+      is_linked? || libra_id.present? || case_urn.present?
+    end
+
+    def linked_to_common_platform?
+      libra_id.to_s.start_with?(COMMON_PLATFORM_PREFIX) && case_urn.present?
+    end
+
+    def linked_case_urn
+      case_urn
     end
 
   private
--- a/app/services/process_xhibit_cases.rb
+++ b/app/services/process_xhibit_cases.rb
@@ -29,10 +29,20 @@ private
   end
 
   def handle_success(response, xhibit_case)
-    return handle_existing_link(xhibit_case) unless response.no_existing_link?
+    if response.existing_link?
+      xhibit_case.action_required!
+      record_error(xhibit_case, :maat, message: existing_link_message(response))
+      return
+    end
 
     defendant_summary = fetch_defendant_summary(xhibit_case)
-    return handle_not_on_common_platform(xhibit_case) if defendant_summary.nil?
+
+    if defendant_summary.nil?
+      # Handle not on common_platform
+      xhibit_case.action_required!
+      record_error(xhibit_case, :common_platform, message: "Case not found on Common Platform")
+      return
+    end
 
     LinkXhibitCase.call(response, xhibit_case, defendant_summary)
   rescue ActiveRecord::RecordInvalid => e
@@ -69,14 +79,12 @@ private
     record_error(xhibit_case, :maat, message: "MAAT application not found")
   end
 
-  def handle_existing_link(xhibit_case)
-    xhibit_case.action_required!
-    record_error(xhibit_case, :maat, message: "MAAT application is already linked")
-  end
-
-  def handle_not_on_common_platform(xhibit_case)
-    xhibit_case.action_required!
-    record_error(xhibit_case, :common_platform, message: "Case not found on Common Platform")
+  def existing_link_message(response)
+    if response.linked_to_common_platform?
+      "MAAT ID already linked with other CP case (#{response.linked_case_urn})"
+    else
+      "MAAT application is already linked"
+    end
   end
 
   # Errors are keyed by step and merged, so that a failure in one step does not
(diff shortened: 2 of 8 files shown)


app/models/maat_api/search_response.rb
```diff
@@ -24,8 +26,16 @@ def maat_id
       response["maatId"]
     end
 
-    def no_existing_link?
-      !is_linked? && !libra_id && !case_urn
+    def existing_link?
+      is_linked? || libra_id.present? || case_urn.present?
+    end
```
alexdesi, line 31: I used direct logic,
so we can use
`if existing_link?`

instead of making our lives miserable with 
`unless no_existing_link?`;)

pezholio: Yeah, that makes sense!

spec/services/process_xhibit_cases_spec.rb
```diff
@@ -149,6 +149,33 @@
         )
       end
     end
+
+    context "when the MAAT application is already linked to a Common Platform case" do
+      let(:cassette) { "maat_api/search_maat_application_linked_to_cp_case" }
```
pezholio, line 154: I think it'd be good for us to look at using Webmock in favour of VCR cassettes, as @sophiebrown01 points out here https://mojdt.slack.com/archives/C076MS401D5/p1788774264045999 - not necessarily a blocker for this PR, but worth thinking about

pezholio (approved):
Looks really good 👍

app/models/maat_api/search_response.rb
sophiebrown01: I for one 100% appreciate this, thanks! 💯

app/models/maat_api/search_response.rb
```diff
@@ -24,8 +26,16 @@ def maat_id
       response["maatId"]
     end
 
-    def no_existing_link?
-      !is_linked? && !libra_id && !case_urn
+    def existing_link?
+      is_linked? || libra_id.present? || case_urn.present?
```
sophiebrown01, line 30: One question: is it possible that `is_linked?` could return `false` but there then be a Libra id or case URN present? If so, how would this not then realistically be already linked?

app/services/process_xhibit_cases.rb
```diff
@@ -69,14 +79,12 @@ def handle_not_found(xhibit_case)
...
-    xhibit_case.action_required!
-    record_error(xhibit_case, :maat, message: "MAAT application is already linked")
-  end
-
-  def handle_not_on_common_platform(xhibit_case)
-    xhibit_case.action_required!
-    record_error(xhibit_case, :common_platform, message: "Case not found on Common Platform")
+  def existing_link_message(response)
+    if response.linked_to_common_platform?
+      "MAAT ID already linked with other CP case (#{response.linked_case_urn})"
```
sophiebrown01, line 84: Just to check - this is new wording? We may want to run it by our content designer?

pezholio: I wonder if we could store this in translations, so any changes are easy to make in the future?

app/models/maat_api/search_response.rb
alexdesi: good point @sophiebrown01  let me double-check this logic, I believe it was in the requirement somewhere 🤔

alexdesi: so, this is the condition in the https://dsdmoj.atlassian.net/wiki/spaces/ACD2/pages/6092227492/DLRM+Implementation+Analysis#Scenario-3%3A-MAAT-record-found%2C-no-existing-link

```
Scenario 3: MAAT record found, no existing link
Condition: libraId is null AND caseUrn is null (with isLinked = false).
```

Also, these data come from the `search-maat-application` we trust them to be consistent,
but to be 100% we must ask Maat Assess for some respoonse body examples

app/services/process_xhibit_cases.rb
alexdesi: @sophiebrown01 the wording are from the ACs in the ticket:

```
Scenario: Store reason for manual action
  Given a MAAT application is already linked to a Common Platform case
...
  Then the error_message field is set to "MAAT ID already linked with other CP case (01AB1234567)"
```

> We may want to run it by our content designer? 
Maybe, it something we can think about at a later stage. We can ask them to review all the messages/errors.

spec/services/process_xhibit_cases_spec.rb
alexdesi: Thanks for sharing the thread @pezholio , I missed it.
Yes, WebMocks are easier to edit. I'll consider it in my next PR, see if it worth changing the cassette withs webmocks.

fyi @sophiebrown01 @leandroalemao

app/services/process_xhibit_cases.rb
alexdesi: it is in the ACs:
```
Scenario: Store reason for manual action
  Given a MAAT application is already linked to a Common Platform case
  ...
  Then the error_message field is set to "MAAT ID already linked with other CP case (01AB1234567)"
```

sophiebrown01: Ah ok, I missed it in the ACs. We should perhaps double-check this as I think those ACs were written by Chuan but didn't go through any sort of content review.

PR #1602, opened 2026-09-08
Update to next-gen CircleCI convenience images

pezholio:
We’re currently having an issue deploying. Seeing an error when running `apt-get update`. Not entirely sure if this will help, but we probably should be using the new images anyway.


pezholio:
Ah, no, this was a red herring, and I updated the orbs, not the Docker images (most of which ARE next-gen anyway)

PR #1604, opened 2026-09-09
[ACD-1435] Improve OpenSearch logging for Common Platform errors

leandroalemao:
#### Ticket

- [LAACONNECT-1435](https://dsdmoj.atlassian.net/browse/LAACONNECT-1435)

#### Why

Currently, errors from the Common Platform (CP) API are not logged with sufficient detail to diagnose issues via OpenSearch.

### What 

- Add rails semantic logger gem for better & structured logs;
- CP errors are logged with structured fields: endpoint, HTTP status, error message, request ID;
- Log level is appropriate;
- Logs are queryable in OpenSearch;
- Fix sidekiq deprecation message;
- Skip all aws provider credentials checks in tests;
- Remove HTTP headers from being logged at all;
- Running `rspec` does not print any "noise" in the screen;
- Fix sidekiq 9 deprecation message;

#### Logs before changes on OpenSearch:

image

#### Logs after changes (with HTTP headers):

image

#### Logs after changes (with no HTTP headers):

image

#### Events # dropped a lot as we're filtering out the `StatusController` logs:

image

#### Rspec runs with no "noise" being printed on screen:

image

[LAACONNECT-1435]: https://dsdmoj.atlassian.net/browse/LAACONNECT-1435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/common_platform/connection.rb
```diff
@@ -10,14 +10,39 @@ class Connection
...
+      def endpoint_for(url)
+        url.respond_to?(:path) ? url.path.to_s.gsub(UUID_PATTERN, ":id") : url.to_s
+      end
+    end
+
     def initialize
       @connection = Faraday.new HOST, options do |connection|
         connection.request :retry, retry_options
         connection.request :json
         connection.response :logger, TaggedLogger, { headers: false, formatter: LogFormatter } do |logger|
```
pezholio: Would it be worth addding the full namespace for `LogFormatter` here? This tripped me up a bit, because I had to hunt around for where it was defined now

app/services/common_platform/connection.rb
```diff
@@ -10,14 +10,39 @@ class Connection
...
-          logger.filter(/(defendantName=)([^&]+)/, '\1[FILTERED]')
-          logger.filter(/(defendantDOB=)([^&]+)/, '\1[FILTERED]')
-          logger.filter(/(defendantNINO=)([^&]+)/, '\1[FILTERED]')
+          PII_FILTERS.each { |pattern, replacement| logger.filter(pattern, replacement) }
```
pezholio, line 45: This is great 👍

pezholio (approved):
This is really cool. Happy with this assuming it all works well in test 👍

app/services/common_platform/connection/failure_middleware.rb
```diff
@@ -0,0 +1,26 @@
...
+          url: Connection.filter_pii(env.url),
+          error_class: e.class.name,
+          error_message: e.message,
+        )
+
+        raise CommonPlatform::Api::Errors::FailedDependency, e
+      end
+    end
+  end
+end
```
alexdesi, line 26: That's a nice class :)
just one thing: can you check if this capture the timeout errors (which are the most frequent) and ssl errors?

app/services/common_platform/connection/log_formatter.rb
```diff
@@ -0,0 +1,59 @@
...
+        )
+      end
+
+      def log_level_for(status)
+        EXPECTED_FAILURE_STATUSES.include?(status) ? :warn : :error
+      end
+
+      def error_message_for(env)
+        # In case of error, Common Platform API returns an HTML page.
+        body = ActionView::Base.full_sanitizer.sanitize(env.body.to_s).strip
```
alexdesi, line 53: it good to escape in case of html,
but most of the time it's a json, correct? 
but there is not a check for the format 🤔

alexdesi (approved):
Nice one @leandroalemao 🚀

alexdesi:
@leandroalemao  I agree with you on removing all those HTTP_* fields that generate lots of noise,
but possibly these two are useful:

- User-Agent tells you which system made a request.
- Client IP. X-Forwarded-For helps when investigating suspicious or unexpected traffic.

we do not need to reinstate the lograge gem, 

```
payload[:user_agent] = request.user_agent
payload[:remote_ip] = request.remote_ip
```
    
should be enough

PR #1605, opened 2026-09-09
Add support for isCivil flag on hearing resulted messages

sophiebrown01:
This PR adds support for the existing `isCivil` flag that can be part of hearing resulted messages received from HMCTS. This flag has never needed to be supported before as CDA has only ever dealt with criminal cases, however with civil proceedings there is now a need to differentiate between criminal and civil cases.

The `isCivil` flag will only ever appear on civil cases and will not even be a property on a prosecution case object for criminal cases.

[Link to story](https://dsdmoj.atlassian.net/browse/LAACONNECT-1521)

### DIFF
--- a/app/models/maat_api/prosecution_case.rb
+++ b/app/models/maat_api/prosecution_case.rb
@@ -1,12 +1,13 @@
 module MaatApi
   class ProsecutionCase
-    attr_reader :hearing_resulted, :hmcts_common_platform_defendant, :maat_reference, :case_urn
+    attr_reader :hearing_resulted, :hmcts_common_platform_defendant, :maat_reference, :case_urn, :is_civil
 
-    def initialize(hearing_resulted, case_urn, defendant, maat_reference)
+    def initialize(hearing_resulted, case_urn, defendant, maat_reference, is_civil)
       @hearing_resulted = hearing_resulted
       @case_urn = case_urn
       @hmcts_common_platform_defendant = defendant
       @maat_reference = maat_reference
+      @is_civil = is_civil || false
     end
 
     def hearing_id
--- a/app/models/hmcts_common_platform/prosecution_case.rb
+++ b/app/models/hmcts_common_platform/prosecution_case.rb
@@ -28,6 +28,10 @@ module HmctsCommonPlatform
       data[:caseStatus]
     end
 
+    def is_civil
+      data[:isCivil]
+    end
+
     def statement_of_facts
       data[:statementOfFacts]
     end
@@ -57,6 +61,7 @@ module HmctsCommonPlatform
         pc.id id
         pc.prosecution_case_identifier prosecution_case_identifier.to_json
         pc.status status
+        pc.is_civil is_civil
         pc.statement_of_facts statement_of_facts
         pc.statement_of_facts_welsh statement_of_facts_welsh
         pc.defendants defendants.map(&:to_json)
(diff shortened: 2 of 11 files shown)


app/models/maat_api/prosecution_case.rb
```diff
@@ -1,12 +1,13 @@
...
-    attr_reader :hearing_resulted, :hmcts_common_platform_defendant, :maat_reference, :case_urn
+    attr_reader :hearing_resulted, :hmcts_common_platform_defendant, :maat_reference, :case_urn, :is_civil
 
-    def initialize(hearing_resulted, case_urn, defendant, maat_reference)
+    def initialize(hearing_resulted, case_urn, defendant, maat_reference, is_civil)
       @hearing_resulted = hearing_resulted
       @case_urn = case_urn
       @hmcts_common_platform_defendant = defendant
       @maat_reference = maat_reference
+      @is_civil = is_civil || false
```
sophiebrown01, line 10: Decided to add the fallback here in the constructor rather than doing this every time in a separate method in this class

leandroalemao (approved):
Hi @sophiebrown01 lgtm.. altho I'd imagine `isCivil` would be translated to just `civil` either in the method or in the new `JBuilder` instance. Maybe it's just me. All good 👍 😄 

```
def civil
  data[:isCivil]
end
```

sophiebrown01:
> Hi Sophie Brown (@sophiebrown01) lgtm.. altho I'd imagine `isCivil` would be translated to just `civil` either in the method or in the new `JBuilder` instance. Maybe it's just me. All good 👍 😄
> 
> ```
> def civil
>   data[:isCivil]
> end
> ```

I did think about the naming of this last night... I'm thinking I might change it to be something more like `is_civil_case?`

leandroalemao:
> > Hi Sophie Brown (Sophie Brown (@sophiebrown01)) lgtm.. altho I'd imagine `isCivil` would be translated to just `civil` either in the method or in the new `JBuilder` instance. Maybe it's just me. All good 👍 😄
> > ```
> > def civil
> >   data[:isCivil]
> > end
> > ```
> 
> I did think about the naming of this last night... I'm thinking I might change it to be something more like `is_civil_case?`

Hi @sophiebrown01 I've seen `is_civil` is a method in `app/models/maat_api/court_application.rb` as well so I guess `is_civil_case` looks better than `is_civil`.. but I'd always try and drop the `is_` or `has_` from method names which returns a boolean flag.. think `civil_case` would be more the idiomatic Ruby on Rails way. Happy to go for the majority of devs opinions. 👍 😄

sophiebrown01:
> > > Hi Sophie Brown (Sophie Brown (Sophie Brown (@sophiebrown01))) lgtm.. altho I'd imagine `isCivil` would be translated to just `civil` either in the method or in the new `JBuilder` instance. Maybe it's just me. All good 👍 😄
> > > ```
> > > def civil
> > >   data[:isCivil]
> > > end
> > > ```
> > 
> > 
> > I did think about the naming of this last night... I'm thinking I might change it to be something more like `is_civil_case?`
> 
> Hi Sophie Brown (@sophiebrown01) I've seen `is_civil` is a method in `app/models/maat_api/court_application.rb` as well so I guess `is_civil_case` looks better than `is_civil`.. but I'd always try and drop the `is_` or `has_` from method names which returns a boolean flag.. think `civil_case` would be more the idiomatic Ruby on Rails way. Happy to go for the majority of devs opinions. 👍 😄

You're right, apologies, my head is still in Java mode. We don't use these prefixes for Ruby ☺️

PR #1606, opened 2026-09-09
[LAACONNECT-1258] Amend error messaging to be more specific for cases already linked to MLRA

pezholio:
https://dsdmoj.atlassian.net/jira/software/c/projects/LAACONNECT/boards/7817?selectedIssue=LAACONNECT-1258

## Problem

When a MAATid is already taken, a case cannot be linked and the error message does not indicate this is because the MAATid is already taken. 

## Background

The validation of requests is handled by contract classes in `app/contracts`, which use [`dry-validate`](https://github.com/dry-rb/dry-validation) under the hood. If the contracts return an invalid response, this is picked up by [`Errors::ContractError`](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/main/app/models/errors.rb) which generates error codes to return back to the client. These error codes are based on the key of the field that triggered the error, so an invalid MAAT ID will always return `maat_reference_contract_failurre`. In VCD, we then look up the error code in the [translation file](https://github.com/ministryofjustice/laa-court-data-ui/blob/98a1e291369b4aa2814c554156eb21832c6f6f51/config/locales/en/cda_errors.yml). As there's no way to differentiate between the types of failure, we just return a generic response.

## Solution

To get around this, we update the ContractError class to check for any metadata when generating the error code. In our contract class, we can then check the body of the error response, allowing us to return a more specific error code. However, this does mean that if anything changes upstream on the MAAT API side, the error message will change. I did consider setting up Pact tests for this, but it's turning out to be quite fiddly, so will look at implementing an E2E test instead that hits the real MAAT API.

## Checklist

The following things need to be done before this is ready to merge (assuming we agree this is the way forward):

- [ ] Add a translation to VCD to show an error when the MAAT ID is in use elsewhere
- [ ] Add a lightweight E2E test to act as a "canary test" if something changes


spec/requests/api/internal/v2/court_application_laa_references_spec.rb
```diff
@@ -125,6 +126,43 @@
...
+        around do |example|
+          VCR.use_cassette("maat_api/maat_reference_already_linked", tag: :maat_api) do
+            example.run
+          end
+        end
+
+        it "renders a JSON response with an unprocessable_content error and a maat_reference_already_linked error code" do
+          post api_internal_v2_court_application_laa_references_path, params: laa_reference, headers: { "Authorization" => "Bearer #{token.token}" }
+
+          expect(response.body).to include("is already linked to a case")
```
alexdesi, line 144: Should we tell what is the case it is already linked to?

app/contracts/concerns/validates_maat_reference.rb
```diff
@@ -0,0 +1,23 @@
+module ValidatesMaatReference
+  extend ActiveSupport::Concern
+  ALREADY_LINKED_MESSAGE = "is already linked to a case".freeze
```
alexdesi: Is this message coming from a common platform?

We should have a convention for error messages from ~~CP~~ MAAT ( to avoid confusion. Maybe we can just add "FROM_MAAT":

`ALREADY_LINKED_MESSAGE_FROM_MAAT`

spec/requests/api/internal/v2/court_application_laa_references_spec.rb
alexdesi: maybe we do not have that info, not sure 🤔

app/contracts/concerns/validates_maat_reference.rb
```diff
@@ -0,0 +1,23 @@
...
+  included do
+    option :maat_reference_validator, default: -> { MaatApi::MaatReferenceValidator }
+
+    rule(:maat_reference) do
+      next unless value
+
+      validation = maat_reference_validator.call(maat_reference: value)
+      next unless validation && validation.status != 200
+
+      key.failure(text: validation.body["message"], code: maat_reference_error_code(validation.body["message"]))
```
alexdesi: I think you're changing the format of the response here: before it was just a string, now it's a {text:, code:} hash, unless I'm misreading it.

Not necessarily bad, but it could break callers (and there's a comment about "backwards compatibility" in Errors::ContractError).
Also, from the code it looks like code: nil is possible, which isn't much help to the client.

app/contracts/concerns/validates_maat_reference.rb
```diff
@@ -0,0 +1,23 @@
...
+  ALREADY_LINKED_MESSAGE = "is already linked to a case".freeze
+
+  included do
+    option :maat_reference_validator, default: -> { MaatApi::MaatReferenceValidator }
+
+    rule(:maat_reference) do
+      next unless value
+
+      validation = maat_reference_validator.call(maat_reference: value)
+      next unless validation && validation.status != 200
```
leandroalemao: Hi @pezholio maybe if we use the opposite in case a non 200 success message is returned? (ie: 204)

```next if validation.nil? || validation.success?```

File: app/contracts/concerns/validates_maat_reference.rb (hunk 3 above)
Comment (leandroalemao) [reply app/contracts/concerns/validates_maat_reference.rb]: Hi @alexdesi agree and also we're calling `validation.body["message"]` twice here.

File: app/contracts/concerns/validates_maat_reference.rb [hunk 5]
```diff
@@ -0,0 +1,23 @@
+module ValidatesMaatReference
+  extend ActiveSupport::Concern
+  ALREADY_LINKED_MESSAGE = "is already linked to a case".freeze
+
+  included do
+    option :maat_reference_validator, default: -> { MaatApi::MaatReferenceValidator }
+
+    rule(:maat_reference) do
```
leandroalemao, line 8: Hi @pezholio genuine question.. I haven't tested.. but do we need the `rule` to be inside the `included do` block? Are we maybe re-defining this method multiple times wherever `include ValidatesMaatReference` is then added?

spec/support/shared_examples/maat_validator.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.shared_examples "a contract that validates maat_reference" do
```
leandroalemao, line 1: Hi @pezholio what do you think about moving this to its own folder as it's shared example file (not a validator)?

ie:
`spec/support/shared_examples/validates_maat_reference.rb`

spec/support/shared_examples/maat_validator.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    before do
+      stub_maat_validation("already_linked_maat_reference", status: 400)
+    end
+
+    it { is_expected.not_to be_a_success }
+    it { is_expected.to have_contract_error("5635423 is already linked to a case.") }
+    it { is_expected.to have_contract_metadata({ code: :maat_reference_already_linked }) }
+
+    context "when the maat api validator is not available" do
```
leandroalemao: Hi @pezholio I was reading this scenario ... so in case `maat_reference` is already linked and the maat api validator is not available? Maybe if we move this to its own context outside `when maat_reference is already linked`?

Also could we just do maybe (I haven't tested it tho)? 
`stub_maat_validation("nil_response", status: 503)`

spec/support/matchers/have_contract_metadata.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require "rspec/expectations"
+
+# have_contract_metadata
+# e.g.
+# expect(contract.call(my_integer: '1')).to have_contract_metadata({ code: :my_integer_not_an_integer })
+#
+RSpec::Matchers.define :have_contract_metadata do |meta|
+  match do |fullfillment|
```
leandroalemao: Hi @pezholio Is it `fullfillment` or `fulfilment`? not sure

leandroalemao (commented):
Hi @pezholio all good.. just left a few comments. Agree with @alexdesi comments as well.. tksss

spec/requests/api/internal/v2/court_application_laa_references_spec.rb
pezholio: We _could_ do it by parsing the error message, but it's a bit brittle, and would require a bit more work on the VCD side of things. Wondering if it might be easier for the error messages to be parsed by CDA and send to VCD wholesale....

app/contracts/concerns/validates_maat_reference.rb
pezholio: > I think you're changing the format of the response here: before it was just a string, now it's a {text:, code:} hash, unless I'm misreading it.
> Not necessarily bad, but it could break callers (and there's a comment about "backwards compatibility" in Errors::ContractError).

Yeah, that's a good point - do we have any consumers of this endpoint other than VCD though? We could probably get rid of this backwards compatibility if so

> Also, from the code it looks like code: nil is possible, which isn't much help to the client.

This is fine, because if `code` is `nil`, we fall back to just using the key as the error code.

app/contracts/concerns/validates_maat_reference.rb
pezholio: Yep, because the call to `rule` essentially happens on the class itself, not on an instance of the class

spec/support/matchers/have_contract_metadata.rb
pezholio: Yeah, I just copied this from another matcher, where it was misspelled. Will fix in both places

spec/support/shared_examples/maat_validator.rb
pezholio: Keen eyes! Good shout - have fixed. As far as the Webmock stubbing goes, I think it's better to stub the response we get back from `CourtApplicationLinkValidator#call`, as this would result in `nil` before anything happens at the HTTP side - for example if the `MAAT_API_API_URL` was missing for any reason.

app/contracts/concerns/validates_maat_reference.rb
pezholio: I've tweaked this a bit now, so ALL errors return `text` and a `code` by way of translations. Now when we return an error code, we return `#{key}_#{error}_contract_failure` as the code, so, as well as knowing what field has caused the failure, we also know WHY it failed, which will allow us to provide multiple translations.

I'm not sure how widely used these codes are outside of the MAAT ID stuff. There only seem to be a handful of translations inside the UI repo.

leandroalemao (approved):
Hi @pezholio lgtm 👍 tkssss 👏

leandroalemao (approved):

PR #1607, opened 2026-09-09
Refactor: convert VCR cassettes to webmock for the maat search api.

alexdesi:
## What
Refactor: 
as discussed on slack, 
the response of MAAT search api is more convenient in webmock (plain json) format rather than VCR cassettes (yaml with full response stored)

I've converted the cassettes to WebMock


pezholio (approved):
Lovely stuff - thanks for this! 🥇

leandroalemao (approved):

PR #1608, opened 2026-09-11
Acd 1359 Implement auto-linking scenario for MAAT application linked to a Libra case

alexdesi:
## What

Implement auto-linking scenario for MAAT application linked to a Libra case

https://dsdmoj.atlassian.net/browse/LAACONNECT-1359

---
I implemented the Scenario 1: MAAT record found, linked to a Libra case Preview).
I also cleaned up the code by defining a `link_state` to clearly differentiate the scenarios, which should make it easier to understand.

### DIFF
--- a/app/models/maat_api/search_response.rb
+++ b/app/models/maat_api/search_response.rb
@@ -34,8 +34,20 @@ module MaatApi
       libra_id.to_s.start_with?(COMMON_PLATFORM_PREFIX) && case_urn.present?
     end
 
-    def linked_case_urn
-      case_urn
+    def linked_to_libra?
+      libra_id.present? && !libra_id.start_with?(COMMON_PLATFORM_PREFIX)
+    end
+
+    def case_urn
+      linking_detail["caseUrn"]
+    end
+
+    def link_state
+      return :unlinked unless existing_link?
+      return :linked_to_common_platform if linked_to_common_platform?
+      return :linked_to_libra if is_linked? && linked_to_libra?
+
+      :link_state_unknown
     end
 
   private
@@ -44,10 +56,6 @@ module MaatApi
 
     def libra_id
       linking_detail["libraId"]
-    end
-
-    def case_urn
-      linking_detail["caseUrn"]
     end
 
     def is_linked?
--- a/app/services/process_xhibit_cases.rb
+++ b/app/services/process_xhibit_cases.rb
@@ -29,12 +29,22 @@ private
   end
 
   def handle_success(response, xhibit_case)
-    if response.existing_link?
-      xhibit_case.action_required!
-      record_error(xhibit_case, :maat, message: existing_link_message(response))
-      return
+    case response.link_state
+    when :linked_to_common_platform
+      flag_manual_action_required(
+        xhibit_case, "MAAT ID already linked with other CP case (#{response.case_urn})"
+      )
+    when :link_state_unknown
+      flag_manual_action_required(xhibit_case, "MAAT link status could not be determined")
+    when :linked_to_libra
+      RequestLibraUnlink.call(response, xhibit_case)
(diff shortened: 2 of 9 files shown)


app/services/process_xhibit_cases.rb
```diff
@@ -29,12 +29,22 @@ def maat_search(xhibit_case)
   end
 
   def handle_success(response, xhibit_case)
```
alexdesi, line 31: I refactored this, the goal was to make it easier to understand if the code performs the operations described in the scenarios in the tickets/confluence

pezholio: Oooh, this is much nicer 👍

app/models/maat_api/search_response.rb
```diff
@@ -34,8 +34,20 @@ def linked_to_common_platform?
...
+    end
+
+    def case_urn
+      linking_detail["caseUrn"]
+    end
+
+    def link_state
+      return :unlinked unless existing_link?
+      return :linked_to_common_platform if linked_to_common_platform?
+      return :linked_to_libra if is_linked? && linked_to_libra?
```
pezholio, line 48: I wonder if `existing_link?`, `linked_to_common_platform?` and `linked_to_libra?` should be private methods now and `link_state` should be our primary method of finding out the state

app/services/request_libra_unlink.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class RequestLibraUnlink < ApplicationService
```
pezholio, line 3: I like this tiny service!

app/services/process_xhibit_cases.rb
```diff
@@ -29,12 +29,22 @@ def maat_search(xhibit_case)
...
-      return
+    case response.link_state
+    when :linked_to_common_platform
+      flag_manual_action_required(
+        xhibit_case, "MAAT ID already linked with other CP case (#{response.case_urn})"
+      )
+    when :link_state_unknown
+      flag_manual_action_required(xhibit_case, "MAAT link status could not be determined")
+    when :linked_to_libra
+      RequestLibraUnlink.call(response, xhibit_case)
```
pezholio, line 40: Might be worth calling out that we're not 100% sure if this will cause us problems if we link to Common Platform BEFORE the request is acted on

pezholio (approved):
I think this all works really well now - just a couple of minor suggestions

app/services/process_xhibit_cases.rb
alexdesi: good point, I think we could do something to improve % or success,
maybe adding sleep(n sec) 🤔

leandroalemao: Hi @pezholio yeah I was thinking about this as well.. we're publishing a message to SQS to unlink from Libra and then linking to CP. What happens to that message is there any consumer waiting for it? what happens if we try and link to CP without unlinking to Libra first?

pezholio: Yeah, me and @sophiebrown01 have been chatting about that. There is _potentially_ some functionality in the MAAT API that will let us check on the status of a link/unlink request, but this will add quite a bit of complexity, so we've decided to keep it simple for now. If there are any issues when we test this in anger we can look again.

app/services/process_xhibit_cases.rb
leandroalemao: Agree 👍 😄 
Just a comment .. is it ok to just [update columns in the db here (bypasses validations)](https://github.com/ministryofjustice/laa-court-data-adaptor/blob/1c83ded3c3186c049f36334b64c0da4fdbc6e62a/app/services/process_xhibit_cases.rb#L98)?

PR #1612, opened 2026-09-16
[LAACONNECT-1527] Fix Sentry events `CommonPlatform::Api::Errors::FailedDependency

leandroalemao:
## What

[LAACONNECT-1527](https://dsdmoj.atlassian.net/browse/LAACONNECT-1527)

## Problem 

We're getting these errors raised on Sentry.

## Solution

- For some reason CP is not always available (maybe it's restarting or there's a connection failure)
- Retry requests that generated the errors;
- Hide them from being raised on Sentry;

[LAACONNECT-1527]: https://dsdmoj.atlassian.net/browse/LAACONNECT-1527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/initializers/sentry.rb
```diff
@@ -16,5 +16,11 @@
       # error reporting. 100% of errors are reported to Sentry.
       transaction_name.in?(EXCLUDE_PATHS) ? 0.0 : 0.05
     end
+
+    config.inspect_exception_causes_for_exclusion = true
+    config.excluded_exceptions += %w[
+      ActionController::RoutingError
+      EOFError
+    ]
```
pezholio, line 23: I think it might be better if we didn't ignore these exceptions for now, as we will probably want to know if we've retried a couple of times and we still get a non-successfuly response. What do you think?

leandroalemao: Hi @pezholio yeah agree that makes 100% sense. tksss

spec/requests/api/internal/v2/prosecution_cases_spec.rb
```diff
@@ -48,10 +48,29 @@
           run_test!
         end
 
+        context "when search returns no results" do
+          let(:'filter[prosecution_case_reference]') { "id-for-no-results" }
+          let(:cassette_name) { "search_prosecution_case/no_results" }
```
pezholio, line 53: In the spirit of our recent decision to back away from cassettes, I think we should change these tests to use Webmock.

spec/services/common_platform/connection_spec.rb
```diff
@@ -59,7 +59,7 @@
       retry_options = {
         retry_statuses: [409, 429, 500, 502, 504],
         max: 3,
-        interval: 0.05,
+        interval: 3,
```
pezholio: Is there a reason for changing the interval in this commit?

leandroalemao: Hi @pezholio It was just because we have interval set as 3 in main and then I put it back. But now after more thoughts, I slightly amended the interval again. tksss

pezholio (approved):
LGTM 👍

PR #1613, opened 2026-09-16
ACD-1524: Delay the CP link after requesting the LIBRA unlink

alexdesi:
When a MAAT application is still linked to a LIBRA case, we now send the unlink message and then link the case to Common Platform **30 seconds later** in a Sidekiq job (`AutoLinkXhibitCaseWorker`), giving MAAT time to process the unlink first. 

The case stays `pending` while the job is queued, so if the job is ever lost the next `ProcessXhibitCases` run simply picks it up again.


app/services/link_xhibit_case.rb
```diff
@@ -29,14 +29,8 @@ def call
 
 private
 
-  attr_reader :maat_search_response, :xhibit_case, :court_data
-
```
pezholio, line 33: Is there a reason for getting rid of the `attr_reader` here?

app/services/auto_link_xhibit_case.rb
```diff
@@ -0,0 +1,27 @@
+# Attempts to link a migrated XHIBIT case to its MAAT application and leaves the
```
pezholio: This looks like quite a significant refactor (it's good though!), but, for me, it's a bit of a surprise to see a refactor in this commit. Would be easier to review if we had the refactor and the actual work in separate commits (sorry to be a pain!)

app/services/process_xhibit_cases.rb
```diff
@@ -31,76 +34,14 @@ def maat_search(xhibit_case)
...
+      xhibit_case.require_action!(:maat, "MAAT ID already linked with other CP case (#{response.case_urn})")
     when :link_state_unknown
-      flag_manual_action_required(xhibit_case, "MAAT link status could not be determined")
+      xhibit_case.require_action!(:maat, "MAAT link status could not be determined")
     when :linked_to_libra
       RequestLibraUnlink.call(response, xhibit_case)
-      # NOTE: here is a potential problem: what does happen if the message in `sqs_url_unlink` is processed
-      # after the case is linked to CP (see post_laa_reference_to_common_platform) ?
-      link_to_common_platform(response, xhibit_case)
+      AutoLinkXhibitCaseWorker.perform_in(LIBRA_UNLINK_DELAY, xhibit_case.id, response.maat_id)
```
pezholio: Oh nice!

pezholio (commented):
Just a couple of small comments. I like the refactor, but I'm finding it difficult to see what changes have happened inamongst the refactoring work.

app/services/link_xhibit_case.rb
alexdesi: I changed a param, and I just though that attr_reader is not necessary. I find it easy and direct read knowing that those are instance vars (@...) rather then using private methods

Anyway, I am not opinionated, I can revert this if you prefer :) 

see https://www.codewithjason.com/dont-wrap-instance-variables-attr_reader-unless-necessary/

pezholio: Ah, fair. I think I tend to use a private `attr_reader` out of habit. I think that's a good justification actually! Will try and get out of the habit 👍

app/services/auto_link_xhibit_case.rb
alexdesi: I undid it and simplified it. The refactor was not necessary, tbh.

pezholio (approved):
LGTM 👍

app/services/process_xhibit_cases.rb
```diff
@@ -38,29 +57,12 @@ def handle_success(response, xhibit_case)
       flag_manual_action_required(xhibit_case, "MAAT link status could not be determined")
     when :linked_to_libra
       RequestLibraUnlink.call(response, xhibit_case)
-      # NOTE: here is a potential problem: what does happen if the message in `sqs_url_unlink` is processed
-      # after the case is linked to CP (see post_laa_reference_to_common_platform) ?
-      link_to_common_platform(response, xhibit_case)
+      LinkXhibitCaseWorker.perform_in(LIBRA_UNLINK_DELAY, xhibit_case.id, response.maat_id)
     when :unlinked
-      link_to_common_platform(response, xhibit_case)
+      link_to_common_platform(xhibit_case, response.maat_id)
```
sophiebrown01, line 62: Is this going to work? I've checked the spec for `MaatApi::MaatApplicationSearcher` and it can return multiple MAAT applications in the body if multiple exist?

